PR #5, opened 2016-03-18
Better Run Script

JRG-Developer:
The current run script is confusing:

Specifically, it makes assumptions that aren't explicitly mentioned (i.e. requires Homebrew) and assumes a specific directory structure.

This PR helps fix this through better inline comments and explicit OUTPUT_FILE and INPUT_DIR variables.

BTW -- I really like the approach you've taken in generating mocks; I'm still struggling to get Cuckoo Generator to create correct mocks, but I think this is a step in the right direction.

Thanks for sharing and open sourcing this! :+1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,29 +37,47 @@ it, simply add the following line to your test target in your Podfile:
 pod "Cuckoo"
 ```
 
-And add the following `Run script` build phase to your test target's configuration:
+And add the following `Run script` build phase to your test target's `Build Phases`:
 
 ```
-# Find the installed Cuckoo version 
-CUCKOO_VERSION=$(grep '\- Cuckoo' "$PROJECT_DIR/Podfile.lock" | grep -o '[0-9]\{1,\}\.[0-9]\{1,\}\.[0-9]\{1,\}')
+# This script assumes you have home brew installed; see http://brew.sh/
 
-# Used to ignore Xcode's environment
-alias run="env -i PATH=$PATH, HOME=$HOME"
+# Find the installed Cuckoo version
+CUCKOO_VERSION=$(grep '\- Cuckoo' "${PROJECT_DIR}/Podfile.lock" | grep -o '[0-9]\{1,\}\.[0-9]\{1,\}\.[0-9]\{1,\}')
+
+# Ignore Xcode's environment
+alias run="env -i PATH=${PATH}, HOME=${HOME}"
 
 cuckoo runtime --check $CUCKOO_VERSION
 cuckooReturn=$?
 
 if [ $cuckooReturn == 1 ]; then
-    # Update local brew repository and upgrade to latest Cuckoo generator
-    run brew update
-    run brew upgrade SwiftKit/cuckoo/cuckoo
+# If cuckoo generator already installed, update local brew repo and upgrade to latest Cuckoo generator
+run brew update
+run brew upgrade SwiftKit/cuckoo/cuckoo
+
 elif [ $cuckooReturn == 127 ]; then
-    # Update local brew repository and install latest Cuckoo generator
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,29 +37,47 @@ it, simply add the following line to your test target in your Podfile:
...
+# Define output file; change "${PROJECT_NAME}Tests" to your test's root source folder, if it's not the default name
+OUTPUT_FILE="./${PROJECT_NAME}Tests/GeneratedMocks.swift"
+echo "Generated Mocks File = ${OUTPUT_FILE}"
+
+# Define input directory; change "${PROJECT_NAME}" to your project's root source folder, if it's not the default name
+INPUT_DIR="./${PROJECT_NAME}"
+echo "Mocks Input Directory = ${INPUT_DIR}"
+
+# Generate mock files; include as many input files as you'd like to create mocks for
+cuckoo generate --runtime $CUCKOO_VERSION --testable EyeSightEnvironments \
```
TadeasKriz: Please replace `EyeSightEnvironment` with `$(PROJECT_NAME)` property as well, because most of the time that will be the name of the module. The `--testable` will add `@testable import EyeSightEnvironments` into the mocks file.

TadeasKriz:
Thanks! I recently found out that there are `Input files` and `Output files` settings in the `Run script` phase so I would like to update CuckooGenerator to support input/output via that as well.

I would also like to add the script itself into the root of this project, so that it will work like fabric's `run` script (you just run `$PODS_ROOT/Fabric/run` or something). That script would download latest CuckooGenerator binary from the github via HTTP so there will be no need for brew.

What do you think?

TadeasKriz:
You said you are struggling with the generation. What seems to be the problem?

README.md
JRG-Developer: Oops, sorry! Local testing as I went... will update now. :]

TadeasKriz: No problem :)

JRG-Developer:
> I would also like to add the script itself into the root of this project

I really like this idea! However, one issue that I might anticipate would be if someone used nonstandard naming or paths. E.g. named their tests something other than "ProjectNameTests." Perhaps, they could set some variable to be used in this case instead? Perhaps something like `CustomTestProjectPath`?

TadeasKriz:
You would still give the script necessary arguments so that wouldn't be a problem. It would just make it one line `Run script` phase instead of this monster one, so users won't need to know what it does behind the curtains :)

JRG-Developer:
> You said you are struggling with the generation. What seems to be the problem?

After figuring out how the paths work (which is something I tried to address/make easier with this PR), I've been able to get a mock to be generated.

However, it has several syntax errors. Is Cuckoo Generator generating for an older version of Swift? I.e. not Swift 2?

JRG-Developer:
> You would still give the script necessary arguments so that wouldn't be a problem.

Sounds great to me!

TadeasKriz:
> However, it has several syntax errors. Is Cuckoo Generator generating for an older version of Swift? I.e. not Swift 2?

The generator should output Swift 2.1 compatible code. Could you send me what syntax errors do you get?

JRG-Developer:
> Could you send me what syntax errors do you get?

I'll try to reproduce in a dummy project and open as an issue under Cuckoo Generator. Cool?

TadeasKriz:
Awesome, thanks!

TadeasKriz:
:heart:

JRG-Developer:
Cool, thanks :+1:

TadeasKriz:
I just started Slackin on our server so you can join SwiftKit team on Slack through http://swiftkit.tmspark.com/ to chat about your issue :)

TadeasKriz:
There was a but in the latest version of Slackin so it was not working. Fixed now :)

PR #11, opened 2016-06-01
Spring cleanup

FilipDolnik:


Cuckoo.xcodeproj/project.pbxproj
```diff
@@ -245,7 +340,7 @@
...
+			shellScript = "# Find the installed Cuckoo version\nCUCKOO_VERSION=\"0.4.0\"\n\n# Used to ignore Xcode's environment\nalias run=\"env -i PATH=$PATH, HOME=$HOME\"\n\ncuckoo runtime --check $CUCKOO_VERSION\ncuckooReturn=$?\n\nif [ $cuckooReturn == 1 ]; then\n# Update local brew repository and upgrade to latest Cuckoo generator\nrun brew update --verbose\nrun brew upgrade --verbose SwiftKit/cuckoo/cuckoo\nelif [ $cuckooReturn == 127 ]; then\n# Update local brew repository and install latest Cuckoo generator\nrun brew install --verbose SwiftKit/cuckoo/cuckoo\nfi\n\n#cuckoo generate --runtime $CUCKOO_VERSION --output \"$PROJECT_DIR/Tests/GeneratedMocks.swift\" \"$PROJECT_DIR/Tests/TestedProtocol.swift\" \"$PROJECT_DIR/Tests/TestedClass.swift\"";
```
TadeasKriz: 0.4.1

TadeasKriz: You forgot to not commit the commented out `cuckoo generate` command.

Tests/CuckooAPITest.swift
```diff
@@ -59,12 +59,12 @@ class MockeryAPITest: XCTestCase {
         
         XCTAssertEqual(mock.withReturn(), "hello world!")
         
-        var helloWorld: String = ""
+        /*var helloWorld: String = ""
```
TadeasKriz: If commenting out, you have to add a comment why is that.

Source/matching/MatcherFunctions.swift
```diff
@@ -11,14 +11,14 @@ private func compareCalls(count: Int, whenNil: Bool = false, using function: (In
 }
 
 private func describeCallMismatch(calls: [StubCall], description: Description) {
-    description.appendText("was called ").appendValue(calls.count).appendText(" times")
+    description.append("was called ", calls.count, " times")
```
TadeasKriz: Why is this changed? (applies to all uses)

Source/matching/Matcher.swift
```diff
@@ -24,10 +24,6 @@ extension Matcher {
     public var matcher: AnyMatcher<MatchedType> {
         return typeErased()
     }
-    
-    public func describeMismatch(input: MatchedType, to description: Description) {
```
TadeasKriz: Why is it removed?

TadeasKriz:
Travis still not happy.

Source/matching/Matcher.swift
FilipDolnik: It was only moved to FunctionMatcher, AnyMatcher has its own implementation.

TadeasKriz: But that's why it was in the protocol extension. Each time you create a different matcher, you don't need to reimplement it.

FilipDolnik: Not sure if there will be another implementations. But maybe we should consider making one as replacement for AnyMatcher<[StubCall]>, which stands for number of function calls in verify.

TadeasKriz: Why? AnyMatcher is just a type erasure for other matchers basically. And there might be more matchers in the future, so why making it more difficult?

PR #16, opened 2016-06-08
Fix/#12

FilipDolnik:


Tests/CuckooAPITest.swift
```diff
@@ -37,49 +22,72 @@ class CuckooAPITest: XCTestCase {
             when(mock.readWriteProperty.set(anyInt())).then {
                 print($0)
             }
+            when(mock.optionalProperty.set(equalTo(nil))).then { _ in
+                when(mock.optionalProperty.get).thenReturn(nil)
+            }
+            when(mock.optionalProperty.set(anyInt())).then { _ in
```
TadeasKriz: We should test if putting in a value will work, like: `mock.optionalProperty.set(10)`

Tests/CuckooAPITest.swift
```diff
@@ -37,49 +22,72 @@ class CuckooAPITest: XCTestCase {
             when(mock.readWriteProperty.set(anyInt())).then {
                 print($0)
             }
+            when(mock.optionalProperty.set(equalTo(nil))).then { _ in
+                when(mock.optionalProperty.get).thenReturn(nil)
```
TadeasKriz: Wait a minute, you cant do this AFAIK. Stubbing outside of the `stub` block is UB.

Tests/TestedProtocol.swift
```diff
@@ -19,9 +21,11 @@ protocol TestedProtocol {
     
     func withThrows() throws
     
-    func withClosure(closure: String -> Int)
+    func withClosure(closure: String -> Int) -> Int
```
TadeasKriz: This should be `withClosureReturiningInt` and it should be a new method. Do not change methods already being tested.

Tests/CuckooAPITest.swift
```diff
@@ -88,43 +96,71 @@ class CuckooAPITest: XCTestCase {
...
         verify(mock, never()).withThrows()
-        
+        verify(mock).withClosure(anyClosure())
+        verify(mock).withOptionalClosure("hello", closure: anyClosure())
     }
-        
-}
\ No newline at end of file
+    
+    private enum TestError: ErrorType {
```
TadeasKriz: I'm not too fond of moving this enum outside of the functions as each test can define its own. But for now let's leave it as you changed it.

Tests/TestedProtocol.swift
FilipDolnik: If I remember correctly there wasn`t any test for withClosure before.

TadeasKriz: This diff shows otherwise. Maybe it was the one you commented out, which it should not be.

PR #39, opened 2016-06-21
Documentation

FilipDolnik:


README.md
```diff
@@ -62,73 +64,164 @@ ${INPUT_DIR}/FileName3.swift
...
+```Bash
+Carthage/Checkouts/Cuckoo/run
+```
+
+Also don't forget to add the Framework into your project.
+
 ## Usage
 
-For the example, lets say we generate the mock for the following protocol:
+Usage of Cuckoo is similar to [Mockito](http://mockito.org/) and [Hamcrest](http://hamcrest.org/). But there are some differences and limitations caused by generating the mocks and Swift language itself. List of all supported features can be found [here](https://github.com/SwiftKit/CuckooGenerator). You can find complete example in [tests](Tests) concretely in [CuckooAPITest](Tests/CuckooAPITest).
```
TadeasKriz: `concretely` is not right.

README.md
```diff
@@ -62,73 +64,164 @@ ${INPUT_DIR}/FileName3.swift
...
+equalTo<T>(value: T, equalWhen equalityFunction: (T, T) -> Bool)
+
+/// Returns a matcher matching any Int value.
+anyInt()
+
+/// Returns a matcher matching any String value.
+anyString()
+
+/// Returns a matcher matching any T value or nil.
+any<T>(type: T.Type = T.self)
```
TadeasKriz: FYI, we should probably remove the `= T.self` from the runtime, to ask the user for the type for more compile-time security.

README.md
```diff
@@ -62,73 +64,164 @@ ${INPUT_DIR}/FileName3.swift
...
+
+/// Returns a matcher matching any closure.
+anyClosure()
+
+/// Returns a matcher matching any non nil value.
+notNil()
+```
+
+Matching of nil can be achieved with `eq(nil)`.
```
TadeasKriz: We could add a note that when Swift adds conditional conformance, it will work with just `nil`.

TadeasKriz:
👍

PR #59, opened 2016-09-07
Fix #54.

### LINKED ISSUE #54: custom initializers causing errors in generated mock file
following code
```
  init(userName: String, password: String)  {
    self.userName = userName
    self.password = password
    self.sessionId = (self.userName+self.password).hash.description
  }
```
generates this:
```
    override init(userName userName: String, password: String) {
        return manager.call("init(userName:String, password:String)", parameters: (userName: userName, password: password), original: observed.map { o in return { (userName userName: String, password: String) in o.init(userName: userName, password: password) } })
    }
```

removing the double designation doesn't help

FilipDolnik:


Generator/Source/CuckooGeneratorFramework/Generator.swift
```diff
@@ -45,13 +45,15 @@ public struct Generator {
...
             code += ""
             code += "private var observed: \(token.name)?"
-            code += ""
-            code += "\(token.accessibility.sourceName)required\(token.implementation ? " override" : "") init() {"
-            code += "}"
-            code += ""
-            code += "\(token.accessibility.sourceName)required init(spyOn victim: \(token.name)) {"
-            code.nest("observed = victim")
-            code += "}"
+            if (token.children.filter { ($0 as? Method)?.isInit == true }.isEmpty) {
```
TadeasKriz: It would be better to replace it with:
`if !token.children.contains({ ($0 as? Method)?.isInit })`

Also, remember to not put parentheses around conditions.

Generator/Source/CuckooGeneratorFramework/Generator.swift
```diff
@@ -81,13 +83,12 @@ public struct Generator {
...
         guard token.accessibility != .Private else { return }
+        guard !token.isInit else { return }
         
         let override = token is ClassMethod ? "override " : ""
         let parametersSignature = token.parameters.enumerate().map { "\($1.attributes.sourceRepresentation)\($1.labelAndNameAtPosition($0)): \($1.type)" }.joinWithSeparator(", ")
 
-        let parametersSignatureWithoutNames = token.parameters.map {
-                "\($0.attributes.sourceRepresentation)\($0.name): \($0.type)"
-            }.joinWithSeparator(", ")
+        let parametersSignatureWithoutNames = token.parameters.map { "\($0.attributes.sourceRepresentation)\($0.name): \($0.type)" }.joinWithSeparator(", ")
```
TadeasKriz: Why did you modify this?

PR #79, opened 2016-12-14
Added support for “bang” optionals in classes (by converting to quest…

lvdstam:
Hello,
when starting to use Cuckoo on our code, I ran into some compatibility issues. Cuckoo doesn't like all our code :-)

First I added support for vars with the ! instead of ? optional indicator.
Second I noticed that the converter would crash on a var without type that is being initialized inline:
var value = Int(5)
After some investigation it turns out that the tokenizer doesn't now the type of value and crashes the generator. Now at least there is some warning in the generated code. Its not ideal, but I thought I would share it with you to get some feedback on a better implementation.

Regards,

Louis

### DIFF
--- a/Generator/Source/CuckooGeneratorFramework/Tokenizer.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Tokenizer.swift
@@ -94,7 +94,7 @@ public struct Tokenizer {
             
             return InstanceVariable(
                 name: name,
-                type: type!,
+                type: (type ?? "Type could not be inferred by Cuckoo").replacingOccurrences(of: "!", with: "?"),
                 accessibility: accessibility!,
                 setterAccessibility: setterAccessibility,
                 range: range!,
--- a/Tests/ClassTest.swift
+++ b/Tests/ClassTest.swift
@@ -58,6 +58,21 @@ class ClassTest: XCTestCase {
         verify(mock).optionalProperty.set(equal(to: 0))
     }
     
+    func testBangOptionalStringProperty() {
+        var called = false
+        stub(mock) { mock in
+            when(mock.bangOptionalStringProperty.get).thenReturn(nil)
+            when(mock.bangOptionalStringProperty.set(anyString())).then { _ in called = true }
+        }
+
+        mock.bangOptionalStringProperty = "Set it"
+
+        XCTAssertNil(mock.bangOptionalStringProperty)
+        XCTAssertTrue(called)
+        _ = verify(mock).bangOptionalStringProperty.get
+        verify(mock).bangOptionalStringProperty.set(equal(to: "Set it"))
+    }
+
     func testNoReturn() {
         var called = false
         stub(mock) { mock in
--- a/Tests/Source/TestedClass.swift
+++ b/Tests/Source/TestedClass.swift
@@ -17,6 +17,8 @@ class TestedClass {
     
     lazy var optionalProperty: Int? = 0
 
+    var bangOptionalStringProperty: String! = ""
+
     func noReturn() {
     }
     


Generator/Source/CuckooGeneratorFramework/Tokenizer.swift
```diff
@@ -94,7 +94,7 @@ public struct Tokenizer {
             
             return InstanceVariable(
                 name: name,
-                type: type!,
+                type: (type ?? "Type could not be inferred by Cuckoo").replacingOccurrences(of: "!", with: "?"),
```
TadeasKriz: We probably shouldn't use `Type could not be inferred by Cuckoo` as a type. Would it make sense to print it as warning to console and then crash?

TadeasKriz: I'd like to improve the compatibility of `!`, instead of changing all `!` to `?`. Changing all `!` to `?` could make the generated code incompatible. For example if the type was `[Something!]`, the type would be incompatible.

@FilipDolnik Please look into `!` support in runtime part of the framework.

Tests/ClassTest.swift
```diff
@@ -58,6 +58,21 @@ class ClassTest: XCTestCase {
         verify(mock).optionalProperty.set(equal(to: 0))
     }
     
+    func testBangOptionalStringProperty() {
```
TadeasKriz: Thanks for the supplied test!

Generator/Source/CuckooGeneratorFramework/Tokenizer.swift
lvdstam: I would very much be in favor of reporting where the problem (type could not be inferred) is reported to the console and if possible continue tokenizing to catch potential other locations where the same problem exists.

lvdstam: You are right converting ! to ? regardless of the context gets the job done for this simple case.

lvdstam: Another question I have: What is the rationale to use sourcekitten instead of the public domain swift compiler?

TadeasKriz: SourceKitten is just a layer sitting on top of the compiler which parses compiler's output to something we can work with in Swift. I didn't work to reimplement the same functionality, therefore I chose SourceKitten to do the job.

lvdstam: Its weird then that the type inference doesn't seem to work. It looks like a bit of post processing of the AST is still needed. Which I expect the compiler should take care of.
I will have to look deeper to understand the issue better.

lvdstam: Actually a type of [Int!] is not allowed in swift 3 (maybe in earlier versions also). It makes sense to have the bang only as a optional indicator of a top level var/function/...

So my approach may still be ok....

TadeasKriz: You are right, I missed that change, `[Int!]` is indeed no longer supported (I don't use implicitly unwrapped optionals). It could still be `(() -> String!)`, but that would be very unusual.

lvdstam: You are right. Interesting.

lvdstam: On second thought this type does not make sense it should be (() -> String?) or (() -> String) but not (() -> String!). I would say the swift compiler should not allow the type.

TadeasKriz: The type should be compilable, as you can have a function `func hello() -> String! { return nil }` and when you reference it by its name, it'll be `(() -> String!)`. But again, that's something 0.1% will be using (probably even less).

However I found another problem with this solution and that is protocols. When you declare protocol with variable using the implicitly unwrapped optional, the generated class does not conform to the protocol, because it's generated with `Int?` instead of `Int!`. I'm not quite sure how is it possible that this works for classes (I'd think that it shouldn't allow shadowing a field that's `Int!` with a field that's `Int?`).

lvdstam: I consider the ! to be a form of syntactic sugar that the compiler should only allow on class variables. For the other types only ? should be allowed to indicate optionality.

Semantically the (() -> Int!) should mean that the Int although optional is always returned hence making it not optional. It is like saying this nullable is actually a non nullable.

What would semantically be the difference of your hello function when we replace the ! with a ?

I think the shadowing actually indicates that the language designers have similar thoughts...

TadeasKriz: You are correct that it doesn't make sense, but I think it's left there because of Objective-C (compiler tells me: `implicitly unwrapped optionals are only allowed at top level and as function results` if I try to use it differently). I'm not saying people will be using the `-> String!`.

The shadowing works the other way as well (parent has `Int?` and child has `Int!`). However we still have to fix it for protocols as it fails with the property not conforming. 

I'd like us to keep the previous type that's taken from the code, but when this: `var bangOptionalStringProperty: Cuckoo.VerifyProperty<String!>` is being generated, check for the exclamation mark and replace it with `?`. That way the generated property in stub and mock will have the same type as mocked class, but it'll be compilable as the `CuckooVerifyProperty` and `Cuckoo.ToBeStubbedProperty` will no longer use the bang.

What do you think?

lvdstam: I think that shoud be fine. Actually it was my first solution for this issue. But then I found the type inference problem and combined the two :-)

As a workaround we could have changed in our code the vars from ! to ? and adjusted the code where needed.

It is a lot harder to determine the cause of the crash in the generator, so until a proper fix is in place, I would be in favor of emitting some warning in the generated file and not crash...

TadeasKriz: You are correct, we need to take care of the crashing. I'll try to check if there is a way to report crash into Xcode as an error, so it would appear in the build info. That would be awesome.

TadeasKriz:
@lvdstam Hi, thank you for this PR and the discussion. I assigned @FilipDolnik to implement it as we discussed and it's now merged. I'll close this PR and I'd like you to try it out before I release it to CocoaPods. Thanks!

lvdstam:
Looks fine to me. I will wait for the cocoapods update and then try it in our project again.

lvdstam:
Didn't see the update on the cocoapods repo yet....

PR #87, opened 2016-12-27
Added function with constructed default values. Don’t know how to pro…

lvdstam:
Found another problem in our codebase. Its about functions that have a default values for parameters. I'm not quite sure how to test it and if my solution is acceptable for you.

### DIFF
--- a/Generator/Source/CuckooGeneratorFramework/Tokenizer.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Tokenizer.swift
@@ -165,7 +165,7 @@ public struct Tokenizer {
         let filteredParameters = parameters.filter({
             let dictionary = $0 as? [String: SourceKitRepresentable]
             let kind = dictionary?[Key.Kind.rawValue] as? String ?? ""
-            return kind == "source.lang.swift.decl.var.parameter"
+            return kind == Kinds.MethodParameter.rawValue
         })
 
         return zip(parameterLabels, filteredParameters).flatMap(tokenize)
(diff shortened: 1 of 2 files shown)


Generator/Source/CuckooGeneratorFramework/Tokenizer.swift
```diff
@@ -157,13 +157,17 @@ public struct Tokenizer {
...
         parameterLabels = Array(parameterLabels.dropLast())
-        
+
         // Substructure can contain some other informations after the parameters.
-        let filteredParameters = parameters[0..<min(parameterLabels.count, parameters.count)]
-        
+        let filteredParameters = parameters.filter({
+            let dictionary = $0 as? [String: SourceKitRepresentable]
+            let kind = dictionary?[Key.Kind.rawValue] as? String ?? ""
+            return kind == "source.lang.swift.decl.var.parameter"
```
TadeasKriz: The kind name should be with the rest of them in single file and referenced that way (see: https://github.com/SwiftKit/Cuckoo/blob/master/Generator/Source/CuckooGeneratorFramework/Tokens/Kinds.swift).

TadeasKriz: And it's there, just use `Kinds.MethodParameter`

lvdstam: Will do. Any suggestions on how to test this?

lvdstam:
Before continuing with the other issues I submitted, I would like to know if my solutions are acceptable. The three pull requests I submitted already result in some conflicts.

TadeasKriz:
This PR is merged even though it doesn't seem so because I rebased it. (2875ec43bd88c85dac18ab99424c9abe8e8a96b5)

PR #89, opened 2016-12-28
Fix/accesibility problem

lvdstam:
It seems that the setup of the Test target is slightly off (it contains the tests as well as the code under test in the same target). This gives rise to subtle accessibility problems. The access modifier of the containing class and the access modifier of the func, var need to be used to calculate the proper access modifiers. Please check and suggest ways to test it.

### DIFF
--- a/Generator/Source/CuckooGeneratorFramework/Generator.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Generator.swift
@@ -17,24 +17,24 @@ public struct Generator {
     
     public func generate() -> String {
         code.clear()
-        declarations.forEach { generate(for: $0) }
+        declarations.forEach { generate(for: $0, withOuterAccessibility: .Internal) }
         return code.code
     }
     
-    private func generate(for token: Token) {
+    private func generate(for token: Token, withOuterAccessibility outerAccessibility: Accessibility) {
         switch token {
         case let containerToken as ContainerToken:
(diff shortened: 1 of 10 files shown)


Generator/Source/CuckooGeneratorFramework/Generator.swift
```diff
@@ -17,24 +17,24 @@ public struct Generator {
     
     public func generate() -> String {
         code.clear()
-        declarations.forEach { generate(for: $0) }
+        declarations.forEach { generate(for: $0, withOuterAccessibility: .Internal) }
```
TadeasKriz: Why is `.Internal` used here?

lvdstam: Internal is the default access level in swift (when you have no modifier).

lvdstam:
Before continuing with the other issues I submitted, I would like to know if my solutions are acceptable. The three pull requests I submitted already result in some conflicts.

TadeasKriz:
@lvdstam Hi, yes I already merged one of your PR and rebased other one, but github didn't let me rebase master on top of it because of conflicts. I'd like to fix that today and merge it. We're very thankful for your contributions, it really helps to push the project forward.

TadeasKriz:
Merged: b0661e94b220b284c80f388d8e1697bd8a929bda

PR #97, opened 2017-01-19
Throw exception instead of fatalError when possible

XianPalin:
This change is attributed to https://github.com/Quick/Quick/blob/master/Sources/Quick/ErrorUtility.swift

I wasn't able to get the test project setup to verify that tests were still passing


TadeasKriz:
Thanks!

PR #109, opened 2017-02-23
fix `private(set)` accessibility variables

mrgerych:


mrgerych:
@TadeasKriz check this to. it will help with 
```private(set) var variable: Type```

PR #113, opened 2017-03-01
Fix private set

mrgerych:
@TadeasKriz review it plz


MatyasKriz:
This feature seems to be in `master` right now, thanks for the PR anyway!

Issue #124, opened 2017-04-25
Segmentation fault 11 with closure and Alamofire?


arnlen:
Source to test:

```swift
// class ApiTools
// [...]

func callApi(with router: URLRequestConvertible,
            from caller: ApiCompatibleProtocol,
            responseStatusCode: Int,
            _ redirectionClosure: @escaping () -> ()) {
  
  Alamofire.request(router)
      .validate(statusCode: [responseStatusCode])
      .validate(contentType: ["application/json"])
      .responseJSON { response in
          switch response.result {
          case .success:
              let json = response.result.value as? [String: Any]
              caller.onSuccessClosure(json!, redirectionClosure)
              
          case .failure(let error):
              // [...]
          }
  }
}
```

Test code:

```swift
let mock = MockApiTools()
        
stub(mock) { stub in
    when(stub.callApi(with: SessionRouter.signin(parameters) as URLRequestConvertible,
        from: sessionService,
        responseStatusCode: 201,
        {})
    )
}
```

Boom error:

```
CompileSwift normal x86_64 <app_path>/AppTests/SessionServiceTests.swift
cd <app_path>
[...]

PR #125, opened 2017-04-25
README: add precision on custom matchers [ci skip]

arnlen:
Following #124, a few more explanations on custom matchers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ We plan to add a **watchOS 2+** support soon.
 
 ## Cuckoo
 
-### Installation
+### 1. Installation
 
 #### CocoaPods
 Cuckoo runtime is available through [CocoaPods](http://cocoapods.org). To install
@@ -108,7 +108,7 @@ Carthage/Checkouts/Cuckoo/run
 
 Also don't forget to add the Framework into your project.
 
-### Usage
+### 2. Usage
 
 Usage of Cuckoo is similar to [Mockito](http://mockito.org/) and [Hamcrest](http://hamcrest.org/). But there are some differences and limitations caused by generating the mocks and Swift language itself. List of all supported features can be found below. You can find complete example in [tests](Tests).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -257,7 +260,30 @@ These basic values are extended to conform to `Matchable`:
 * `UInt32`
 * `UInt64`
 
-Note: Optional types (for example `Int?`) cannot be used directly. You need to wrap them with `equal(to)` function.
+Note: Optional types (for example `Int?`) cannot be used directly. You need to wrap them with `equal(to)` function (see bellow).
+
+#### B) Custom matchers
+
+If Cuckoo doesn't know to type your try to compare, you have to write your own method `equal(to:)` using a `ParameterMatcher`. Add this this method to your test file:
```
Sparkesix: I think this should be "If Cuckoo doesn't know **the** type you are try**ing** to compare, ..."

Also, there is duplicate "**this**" in the next sentence.

README.md
```diff
@@ -257,7 +260,30 @@ These basic values are extended to conform to `Matchable`:
...
+
+```
+Command failed due to signal: Segmentation fault: 11
+```
+
+For details and implementation example (with Alamofire), see [this issue](https://github.com/Brightify/Cuckoo/issues/124).
+
+#### Parameter matchers
 
 `ParameterMatcher` also conform to `Matchable`. You can create your own `ParameterMatcher` instances or if you want to directly use your custom types there is the `Matchable` protocol. Standard instances of `ParameterMatcher` can be obtain via these functions:
```
Sparkesix: "also conform**s** to ..."

In the last sentence: "can be obtain**ed** ..."

README.md
```diff
@@ -257,7 +260,30 @@ These basic values are extended to conform to `Matchable`:
...
+```swift
+func equal(to value: YourCustomType) -> ParameterMatcher<YourCustomType> {
+    return ParameterMatcher { tested in
+        // Implementation of your equality test.
+        // ie: (try? tested.method()) == (try? value.method())
+    }
+}
+```
+
+⚠️ If you try to match an object with an unknown or non-`Matchable` type, this could lead to a:
```
Sparkesix: This may be subjective, but I feel that "this could lead to a:" would feel more natural as "it can lead to:". Removing the "a" as well, I would suggest keep it if it was one sentence, but when you're using a colon, it seems pointless.

My reasoning is that "this" in this sense is mainly used at the beginning of sentences, but here it's not a new sentence.
Consider these two shorter examples,
"If you fall down, it can lead to a hospital visit."
and
"If you fall down, this can lead to a hospital visit."

arnlen:
Thx a lot @Sparkesix for this review! I'll make the suggested changes. 👍

README.md
arnlen: Done! 👍

README.md
arnlen: Very clear and instructive: thanks a lot for those explanations! 👍 It makes totally sense, and I've updated the doc accordingly.

README.md
arnlen: Done! 👍

README.md
Sparkesix: Happy to help. :)

MatyasKriz:
Merged. Thanks for the PR!

Issue #132, opened 2017-05-27
Unable to generate GeneratedMocks.swift since Source Kitten 0.17.6 release


silencecm:
I'm an Android dev learning swift / ios on my own. I'm using TDD to port my released Android app over to iOS. This is my first attempt using cocoapods.
```
ProductName:	Mac OS X
ProductVersion:	10.12.4
BuildVersion:	16E195
XCode: Version 8.2.1 (8C1002)
Build Target: 9.3
Swift 3?
```
When attempting to generate a mock file I received the following error:
```
/Users/user/Documents/ios/my_project/Pods/Cuckoo/Generator/Packages/SourceKitten.git/Package.swift:17:3: error: argument 'swiftLanguageVersions' must precede argument 'dependencies'
  swiftLanguageVersions: [3, 4],
  ^
Can't parse Package.swift manifest file because it contains invalid format. Fix Package.swift file format and try again.
```
I resolved this issue (for now) by modifying the following file
`/Users/user/Documents/ios/my_project/Pods/Cuckoo/Generator/Package.swift` and updating Source Kitten to use version 16 max:
```
.Package(url: "https://github.com/jpsim/SourceKitten.git", versions: Version(0, 15, 0)..<Version(0, 16, .max))
```
My tests are now building as passing as expected. Although I haven't done any mocking yet I assume this issue needs to be addressed by someone. Source Kitten 0.17.6 was released 5 days ago.

silencecm:
Closing per comment on PR #133

PR #133, opened 2017-06-11
Only use SourceKitten max version 16

silencecm:
Proposed fix for issue #132


TadeasKriz:
Hi, thanks for the commit. What Xcode version are you using? I wasn't able to reproduce #132. Thanks.

silencecm:
I'm using XCode Version 8.2.1 (8C1002). After rolling back this change Cuckoo appears to be working as expected. I must have made an error during initial configuration. Thanks for the assistance with this.

PR #161, opened 2017-12-16
Update README.

vaibhavgohel:
Generate testable command will fail if any of parent directory contain space.


MatyasKriz:
Merged. Thanks for the PR!

PR #169, opened 2018-02-20
Fixes a regression introduced in 96ebbbf

ggrell:
There was an attempt to make the file order deterministic in 96ebbbf, but the sorting that was added only worked if all mocks were generated in a single file. If generated to a path, which creates a file per mock, the filenames and the file contents did not match. Solution was to sort the set of paths passed in as options instead, moving the sorting a little earlier than before.


TadeasKriz:
@ggrell Merged. I'll close this PR as I invalidated it by rebasing your commit. Thank you for your contribution!

PR #170, opened 2018-02-21
The sort of input files occurred to soon

ggrell:
The files input via options are mapped to Path objects then sorted. This actually happened too soon, since the array is then converted to a Set, then converted back to an Array. The Set conversion changes the order (since it is unordered) and those we lost the effect of the sort. I move the sort to the correct place, which is on the outermost Array so that the inputPathValues array is now correctly sorted.


MatyasKriz:
Merged. Thanks for the PR!

PR #190, opened 2018-04-06
Add a debug flag that generates general info above methods when used.

### LINKED ISSUE #167: Unnecessary comment generated preceding a function signature
I am using Cuckoo 0.10.2. It turns out the generated code corresponding to a function of a class/protocol contains some a lengthy comment which, I guess, is the metadata of that function.

Here is the protocol I wish to mock
```swift
protocol ApiClientProtocol {
    func request(_ method: HttpMethod, _ urlString: String, completion: ((Data?, Error?) -> Void)?)
}
```

And this is a generated code. For brevity, I only excerpt a part of it. Please look at the comment `// ["name": "request"...`
```swift
class MockApiClientProtocol: ApiClientProtocol, Cuckoo.Mock {
    ...   
    // ["name": "request", "returnSignature": "", "fullyQualifiedName": "request(_: HttpMethod, _: String, completion: ((Data?, Error?) -> Void)?)", "parameterSignature": "_ method: HttpMethod, _ urlString: String, completion: ((Data?, Error?) -> Void)?", "parameterSignatureWithoutNames": "method: HttpMethod, urlString: String, completion: ((Data?, Error?) -> Void)?", "inputTypes": "HttpMethod, String, ((Data?, Error?) -> Void)?", "isThrowing": false, "isInit": false, "hasClosureParams": true, "@type": "ProtocolMethod", "accessibility": "", "parameterNames": "method, urlString, completion", "call": "method, urlString, completion: completion", "parameters": [CuckooGeneratorFramework.MethodParameter(label: nil, name: "method", type: "HttpMethod", range: CountableRange(261..<281), nameRange: CountableRange(0..<0)), CuckooGeneratorFramework.MethodParameter(label: nil, name: "urlString", type: "String", range: CountableRange(283..<302), nameRange: CountableRange(0..<0)), CuckooGeneratorFramework.MethodParameter(label: Optional("completion"), name: "completion", type: "((Data?, Error?) -> Void)?", range: CountableRange(304..<342), nameRange: CountableRange(304..<314))], "returnType": "Void", "isOptional": false, "stubFunction": "Cuckoo.StubNoReturnFunction"]
    func request(_ method: HttpMethod, _ urlString: String, completion: ((Data?, Error?) -> Void)?)  {
        ...
    }
    ...
}
```

I would argue that such comment is redundant because it only matters to generating code.
So, do we have any option to turn it off (without performing a post-processing)?

Below are my environment:
- Cuckoo: 0.10.2
- XCode: 9.2
- Swift: 4.0

Thank you!

MatyasKriz:
Resolves #167.

### DIFF
--- a/Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
+++ b/Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
@@ -12,16 +12,16 @@ import SourceKittenFramework
 import FileKit
 import CuckooGeneratorFramework
 
-private func curry<P1, P2, P3, P4, P5, P6, P7, P8, P9, R>(_ f: @escaping (P1, P2, P3, P4, P5, P6, P7, P8, P9) -> R)
-    -> (P1) -> (P2) -> (P3) -> (P4) -> (P5) -> (P6) -> (P7) -> (P8) -> (P9) -> R {
-        return { p1 in { p2 in { p3 in { p4 in { p5 in { p6 in { p7 in { p8 in { p9 in f(p1, p2, p3, p4, p5, p6, p7, p8, p9) } } } } } } } } }
+private func curry<P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, R>(_ f: @escaping (P1, P2, P3, P4, P5, P6, P7, P8, P9, P10) -> R)
+    -> (P1) -> (P2) -> (P3) -> (P4) -> (P5) -> (P6) -> (P7) -> (P8) -> (P9) -> (P10) -> R {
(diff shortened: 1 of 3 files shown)


Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
```diff
@@ -123,6 +125,8 @@ public struct GenerateMocksCommand: CommandProtocol {
 
             let noClassMocking: Result<Bool, CommandantError<ClientError>> = m <| Option(key: "no-class-mocking", defaultValue: false, usage: "Do not generate mocks for classes.")
 
+            let debugMode: Result<Bool, CommandantError<ClientError>> = m <| Option(key: "debug", defaultValue: false, usage: "Run generator in debug mode.")
```
TadeasKriz: This could be a `--debug` `Switch`: https://github.com/Carthage/Commandant/blob/master/Sources/Commandant/Switch.swift

PR #191, opened 2018-04-07
Add `glob` switch.

MatyasKriz:
Resolves #171

PR #192, opened 2018-04-07
Add regular expression `class` and `protocol` matching.

MatyasKriz:
Branched from `glob` branch, so this one should be resolved before #191.

### DIFF
--- a/Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
+++ b/Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
@@ -11,13 +11,14 @@ import Result
 import SourceKittenFramework
 import FileKit
 import CuckooGeneratorFramework
+import Foundation
 
-private func curry<P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11, R>
-    (_ f: @escaping (P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11) -> R)
-    -> (P1) -> (P2) -> (P3) -> (P4) -> (P5) -> (P6) -> (P7) -> (P8) -> (P9) -> (P10) -> (P11) -> R {
-        return { p1 in { p2 in { p3 in { p4 in { p5 in { p6 in { p7 in { p8 in { p9 in { p10 in { p11 in
-            f(p1, p2, p3, p4, p5, p6, p7, p8, p9, p10, p11)
-        } } } } } } } } } } }
+private func curry<P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11, P12, R>
+    (_ f: @escaping (P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11, P12) -> R)
+    -> (P1) -> (P2) -> (P3) -> (P4) -> (P5) -> (P6) -> (P7) -> (P8) -> (P9) -> (P10) -> (P11) -> (P12) -> R {
+        return { p1 in { p2 in { p3 in { p4 in { p5 in { p6 in { p7 in { p8 in { p9 in { p10 in { p11 in { p12 in
+            f(p1, p2, p3, p4, p5, p6, p7, p8, p9, p10, p11, p12)
+        } } } } } } } } } } } }
 }
 
 public struct GenerateMocksCommand: CommandProtocol {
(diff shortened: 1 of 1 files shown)


Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
```diff
@@ -11,13 +11,14 @@ import Result
...
-
-private func curry<P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11, R>
-    (_ f: @escaping (P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11) -> R)
-    -> (P1) -> (P2) -> (P3) -> (P4) -> (P5) -> (P6) -> (P7) -> (P8) -> (P9) -> (P10) -> (P11) -> R {
-        return { p1 in { p2 in { p3 in { p4 in { p5 in { p6 in { p7 in { p8 in { p9 in { p10 in { p11 in
-            f(p1, p2, p3, p4, p5, p6, p7, p8, p9, p10, p11)
-        } } } } } } } } } } }
+import Foundation
+
+private func curry<P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11, P12, R>
```
TadeasKriz: This is gross, we need to fix this in the future.

Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
```diff
@@ -40,7 +41,8 @@ public struct GenerateMocksCommand: CommandProtocol {
         let tokensWithInheritance = options.noInheritance ? tokens : mergeInheritance(tokens)
         let tokensWithoutClasses = options.noClassMocking ? removeClasses(tokensWithInheritance) : tokensWithInheritance
         // filter excluded classes/protocols
-        let parsedFiles = removeClassesAndProtocols(from: tokensWithoutClasses, in: options.exclude)
+        let matchedTokens = removeClassesAndProtocols(from: tokensWithoutClasses, using: options.regex)
+        let parsedFiles = removeClassesAndProtocols(from: matchedTokens, in: options.exclude)
```
TadeasKriz: Let's make this iterate less with something like:

```
private func removeTypes(from files: [FileRepresentation], using filters: [(Token) -> Bool)]) {
    // Only keep those that pass all filters
    let filter: (Token) -> Bool = { token in 
        filters.first(where: { !$0(token) }) == null
    }
    return files.flatMap { file in
        let filteredDeclarations = file.declarations.filter(filter)
        guard !filteredDeclarations.isEmpty else { return nil }
        return FileRepresentation(sourceFile: file.sourceFile, declarations: filteredDeclarations)
    }
}

private func ignoreClasses(token: Token) -> Bool {
    return !(token is ClassDeclaration)
}

private func keepMatching(pattern: String) -> (Token) -> Bool {
    do {
        let regex = try NSRegularExpression(pattern: pattern, options: [])

        return { token in 
            guard let containerToken as? containerToken else { return true }
            return regex.firstMatch(in: token.name, options: [], range: NSMakeRange(0, token.name.count)) != nil
        }
    } catch {
        fatalError("Invalid regular expression: " + (error as NSError).description)
    }
}

private func ignoreIfExists(in excluded: [String]) -> (Token) -> Bool {
    let excludedSet = Set(excluded)
    return { token in 
        !excluded.contains(token.name)
    }
}
```

And usage would look like:

```
var typeFilters = [] as [(Token) -> Bool]
if options.noClassMocking {
    typeFilters.append(ignoreClasses)
}
if !options.regex.isEmpty {
    typeFilters.append(keepMatching(pattern: options.regex))
}
if !options.exclude.isEmpty {
    typeFilters.append(ignoreIfExists(in: options.exclude))
}
let parsedFiles = removeTypes(from: tokensWithInheritance, typeFilters)
```

Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
MatyasKriz: Yeah, I don't like it either but I don't understand it well enough to eliminate it.

Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
MatyasKriz: Amused by your use of `null`. 😄

TadeasKriz:
Spent too much time in JavaScript these past days 😀

-- 
Tadeas Kriz

Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
TadeasKriz: Spent too much time in JavaScript these past days 😀

PR #205, opened 2018-06-05
[WIP] Reintroduce support for pre-0.11.0 Cuckoo spies.

TadeasKriz:
This PR reintroduces support for pre-0.11.0 spies, but we're calling them "default implementation" instead of "spy".

## WIP

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Source/Mock/Mock.swift
+++ b/Source/Mock/Mock.swift
@@ -18,6 +18,8 @@ public protocol Mock: HasMockManager {
     func getStubbingProxy() -> Stubbing
     
     func getVerificationProxy(_ callMatcher: CallMatcher, sourceLocation: SourceLocation) -> Verification
+
+    func enableDefaultImplementation(_ stub: MocksType)
 }
 
 public extension Mock {
@@ -27,6 +29,11 @@ public extension Mock {
     
     func getVerificationProxy(_ callMatcher: CallMatcher, sourceLocation: SourceLocation) -> Verification {
         return Verification(manager: cuckoo_manager, callMatcher: callMatcher, sourceLocation: sourceLocation)
+    }
+
+    func withEnabledDefaultImplementation(_ stub: MocksType) -> Self {
+        enableDefaultImplementation(stub)
+        return self
     }
 }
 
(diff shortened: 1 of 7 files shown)


Source/Mock/Mock.swift
```diff
@@ -28,6 +30,11 @@ public extension Mock {
     func getVerificationProxy(_ callMatcher: CallMatcher, sourceLocation: SourceLocation) -> Verification {
         return Verification(manager: cuckoo_manager, callMatcher: callMatcher, sourceLocation: sourceLocation)
     }
+
+    func withEnabledDefaultImplementation(_ stub: MocksType) -> Self {
```
macduy: Is it possible to have the stub argument default to the generated `*Stub` class? From a practical point of view this would make sense for us (as 99% of our mocks are from protocols and are stubbed to the `*Stub` as a starting point), though you may have a different opinion :) What do you think?

macduy:
Thanks for picking this up @TadeasKriz, it works great!

Source/Mock/Mock.swift
TadeasKriz: I'm not sure if that's a good idea as we can't be sure the stubs will have a no-arg initializer.

macduy: Ah that's a good point. Thanks

macduy:
Hi Tadeas, do you know if there are plans to have this merged some time soon?

TadeasKriz:
@macduy Could you take a look at: https://github.com/Brightify/Cuckoo/pull/214 and see if it works in your project? This is something I'd like to merge together with this PR.

macduy:
Will take a look.

macduy:
Hi @TadeasKriz, sorry for the long pause, I got pulled away to do some other work but finally found some time to check this. Good news is, it looks like it works for us:

image

Thanks for making this fix! When do you estimate you could merge this into a new Cuckoo release? We are having a slight issue that every time we run `pod update` we need to revert the binary changes to `cuckoo_generator`.

TadeasKriz:
Hey @macduy, that's great news! I'll try to merge during the upcoming week. Please try to remind me in case I forget. Thanks!

macduy:
@TadeasKriz that would be great, thanks again for coming up with a solution!

macduy:
Ping @TadeasKriz ;)

PR #215, opened 2018-08-09
Add explore utility function to build script

aardouin:
Adds a sample to generate mock from all files inside a folder.


MatyasKriz:
Hey, thanks for the PR!

However, we recently added a glob functionality (#191), does that not fulfill your needs?

If not, we can discuss this matter further, otherwise we would rather have that glob functionality presented in the `README.md` (it's not there yet).

aardouin:
I didn't see the glob feature ( as it was not documented yet ). So yes glob do answer my problem I guess. 
I will give it a try as soon as I get some time to work on it.

PR #232, opened 2018-11-05
Add `@available` attribute support.

### LINKED ISSUE #231: <@available> attribute
I have a class which I would like to mock, the class is assigned with the following attribute @available(iOS 10.0, *).
The issue I see, is that GeneratedMocks.swift file does not assign the attribute to the generated mocks, resulting in an error '<class> us only available on IOS 10.0 or newer'.

MatyasKriz:
Also added some parameters into the `replace(children:)` method, it was leaving out some information.

Also fixes #231.


MatyasKriz:
On second thought, adding `@available` might be more tricky than I thought because it has parameters inside it.

PR #236, opened 2018-11-07
Show class accessibility.

### LINKED ISSUE #218: Generate mocks with public accessibility
In one of my swift files I have a public protocol defined as:

```
public protocol XYZ {
  var abc: String? { get}
}
```

When a mock is generated for this the default accessibility is not public:

```
class MockXYZ {
...
}
```

In my specific project I want my mock to be visible in another target (we have multiple test targets, and I don't want to have to regenerate the mocks for each of the targets). Since the mocks aren't generated as `public` even though the protocol is public, I have to import this as `@testable`. 

As best I can see, there's no way to get around this right now, please let me know if I'm wrong.

MatyasKriz:
It's already enforced in the tokenizer by `!`, so it should have a value and therefore there's nothing wrong with not showing it, is there?

Closes #218.

### DIFF
--- a/Generator/Source/CuckooGeneratorFramework/Tokens/ContainerToken.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Tokens/ContainerToken.swift
@@ -35,7 +35,7 @@ extension ContainerToken {
 
         return [
             "name": name,
-            "accessibility": accessibility,
+            "accessibility": accessibility.sourceName,
             "isAccessible": accessibility.isAccessible,
             "children": children.map { $0.serializeWithType() },
             "properties": properties,
--- a/Generator/Source/CuckooGeneratorFramework/Templates/MockTemplate.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Templates/MockTemplate.swift
@@ -10,7 +10,7 @@ import Foundation
 extension Templates {
     static let mock = """
 {% for container in containers %}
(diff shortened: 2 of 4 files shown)


Generator/Source/CuckooGeneratorFramework/Tokens/ContainerToken.swift
```diff
@@ -35,7 +35,7 @@ extension ContainerToken {
 
         return [
             "name": name,
-            "accessibility": accessibility,
+            "accessibility": accessibility.sourceName,
```
TadeasKriz: Doesn't this break something? Feels like it could break something :D

MatyasKriz: Actually fixes everything. `Accessibility` is a `String Enum` with SourceKit identifiers for some accessibility attributes. It had no place in serializing or Stencil templates, it wasn't used even. I added the `sourceName` to be able to use it in code and started using it in Stencil templates.

jamezilla:
@TadeasKriz I'm looking for a solution to mocking frameworks. Seems like this would allow us to ship a mock framework along side our real framework. Does that seem right? Is there any timeline for merging this?

TadeasKriz:
Hey @jamezilla, thanks for the reminder. I don't know why we didn't merge this one already.

@MatyasKriz could you rebase and merge this PR? Thanks!

PR #237, opened 2018-11-09
Transform IUO into `Optional` to prevent errors.

### LINKED ISSUE #221: ImplicitlyUnwrappedOptionals are being deprecated in Swift 4.2
Hi

We are using Cuckoo 0.11.3 and our project builds in Swift 4.1 and noticed this warning in the generated mocks:

```
Using ‘!’ in this location is deprecated and will be removed in a future release; consider changing this to ‘?’ instead
```

From my understanding, in Swift 4.2 `ImplicitlyUnwrappedOptionals` will no longer be a valid type: https://github.com/apple/swift-evolution/blob/master/proposals/0054-abolish-iuo.md#motivation

This will **break Cuckoo-generated mocks** in cases where mocked protocol or class has a IUO variable:

```swift
protocol StringIdentifier {
    var identifier: String! { get }
}
```

will generate

```swift
class StringIdentifierStub: StringIdentifier {
    
     var identifier: String! {
        get {
            return DefaultValueRegistry.defaultValue(for: (String!).self)
        }
        
        set { }
    }
}
```
The `(String!).self` type gives a warning in Swift 4.1. and will become an error in Swift 4.2. A manual fix would be use a cast to the optional type instead:

```swift
return DefaultValueRegistry.defaultValue(for: (String?).self)
```

Do you think this can achieved? I'm guessing the change could be made here, but I'm not very familiar with Stencil.

https://github.com/Brightify/Cuckoo/blob/0a5c4f7f1d9d56e57e81a9854e5b572ba1186350/Generator/Source/CuckooGeneratorFramework/Templates/NopImplStubTemplate.swift#L14

MatyasKriz:
Fixes #221.

### DIFF
--- a/Generator/Source/CuckooGeneratorFramework/Tokens/InstanceVariable.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Tokens/InstanceVariable.swift
@@ -29,14 +29,13 @@ public struct InstanceVariable: Token {
     }
 
     public func serialize() -> [String : Any] {
+        let readOnlyString = readOnly ? "ReadOnly" : ""
         return [
             "name": name,
             "type": type,
             "accessibility": accessibility.sourceName,
             "isReadOnly": readOnly,
-            "stubType": readOnly ?
-                (overriding ? "ClassToBeStubbedReadOnlyProperty" : "ProtocolToBeStubbedReadOnlyProperty") :
-                (overriding ? "ClassToBeStubbedProperty" : "ProtocolToBeStubbedProperty")
+            "stubType": overriding ? "ClassToBeStubbed\(readOnlyString)Property" : "ProtocolToBeStubbed\(readOnlyString)Property"
         ]
     }
 }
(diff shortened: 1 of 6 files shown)


Generator/Source/CuckooGeneratorFramework/Tokens/InstanceVariable.swift
```diff
@@ -29,14 +29,13 @@ public struct InstanceVariable: Token {
...
+        let readOnlyString = readOnly ? "ReadOnly" : ""
         return [
             "name": name,
             "type": type,
             "accessibility": accessibility.sourceName,
             "isReadOnly": readOnly,
-            "stubType": readOnly ?
-                (overriding ? "ClassToBeStubbedReadOnlyProperty" : "ProtocolToBeStubbedReadOnlyProperty") :
-                (overriding ? "ClassToBeStubbedProperty" : "ProtocolToBeStubbedProperty")
+            "stubType": overriding ? "ClassToBeStubbed\(readOnlyString)Property" : "ProtocolToBeStubbed\(readOnlyString)Property"
```
TadeasKriz: Nice one!

PR #244, opened 2018-11-16
Ignore classes marked as `final`.

### LINKED ISSUE #242: Prevent mocking of final class.
If I have a class defined as `final class Foo {}`, Cuckoo will create a mock for it as such: `class MockFoo: Foo`. This causes the swift-compiler to complain about it attempting to subclass a final class. 

I think Cuckoo should skip classes defined as `final`.

MatyasKriz:
Fixes #242.

### DIFF
--- a/Generator/Source/CuckooGeneratorFramework/Tokenizer.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Tokenizer.swift
@@ -12,9 +12,11 @@ import SourceKittenFramework
 public struct Tokenizer {
     private let file: File
     private let source: String
+    private let debugMode: Bool
 
-    public init(sourceFile: File) {
+    public init(sourceFile: File, debugMode: Bool) {
         self.file = sourceFile
+        self.debugMode = debugMode
 
         source = sourceFile.contents
     }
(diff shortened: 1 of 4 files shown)


TadeasKriz:
Any idea why does the Travis CI build fail?

Tests/Source/TestedClass.swift
```diff
@@ -194,3 +194,7 @@ class ClassUsingInnerEnum {
         return .foo
     }
 }
+
+final class FinalClass {
```
TadeasKriz: Add a "test" for this by adding `class MockFinalClass { }` under this class. This will add another layer of verification (one is verified that the `GeneratedMocks.swift` can be compiled with this input file, the other that there is no `MockFinalClass` class generated).

Generator/Source/CuckooGeneratorFramework/Tokenizer.swift
```diff
@@ -94,6 +94,11 @@ public struct Tokenizer {
                 attributes: attributes)
 
         case Kinds.ClassDeclaration.rawValue:
+            guard !attributes.map({ $0.kind }).contains(.final) else {
+                fputs("Cuckoo: Ignoring mocking of class \(name) because it's marked `final`.\n", stdout)
```
TadeasKriz: Please check if the `-d` option is enabled, otherwise don't log the information. This could be spammy if used with the glob functionality.

Tests/Source/TestedClass.swift
TadeasKriz: And `class FinalClassStub { }` too (forgot about that one).

MatyasKriz: Will work unless the naming convention changes.

MatyasKriz:
I think it's the `Xcode 9`. Should I make it compilable under Swift 4?

TadeasKriz:
If that's easy enough to do

MatyasKriz:
@TadeasKriz, please, another round.

PR #246, opened 2018-11-19
Add `—clean` option to the run script to rebuild the Cuckoo Generator.

MatyasKriz:


TadeasKriz:
This doesn't behave as `clean` should, but I'm okay with it.

PR #248, opened 2018-11-30
Generics support

### LINKED ISSUE #94: Protocol with an associatedtype generates a mock that does not compile
The following fails to generate a proper mock:

protocol SomeProtocol {
    associatedtype SomeAssociatedType
    func someFunc(value: SomeAssociatedType)
}

### LINKED ISSUE #131: Missing Generics / Template
Is there any way to build GeneratedMocks of classes which use templates/generics?

Original method signature:
func save<T>(_ model: T) throws -> T where T:Object

Generated method signature:
public func save(_ model: T)  throws -> T

Then XCode complains at GeneratedMocks.swift code, like: "Use of undeclared type 'T'".

Thanks in advance..

### LINKED ISSUE #206: Handle Mocking protocols with Associated types & generics
I have a protocol I want to mock - problem is it has associated types and generics

something like:

```
protocol MyProtocol {

    associatedtype MyGeneric: Equatable

    func doSomething(with thing: SomeStruct<MyGeneric>)
}
```

When generating a mock for this, Cuckoo generates code that won't compile due to the [classic 'self or associated type' error](https://www.natashatherobot.com/swift-type-erasure/). Code such as this is generated:

```
class MockMyProtocol: MyProtocol, Cuckoo.ProtocolMock {
    typealias MocksType = MockMyProtocol
    private var __defaultImplStub: MockMyProtocol? // <-- this line won't compile, for example
```

Note: I'm generating with something along the lines of:
```./Pods/Cuckoo/run generate --output tmp.swift --no-class-mocking --no-timestamp Path/To/MyProtocol.swift
```

As you can see in the solution section of the blog post above - you need to write/generate a concrete struct - `AnyMyProtocol<T>` in this case.

I'd be happy to in some way mark my protocol with all the information Cuckoo needs to generate working code - (what the boxing type is, what the generics are, even what I'd like them erased to).

Could Cuckoo support this mode of operation?

MatyasKriz:
Closes #131, closes #206, and closes #94.


MatyasKriz:
Hey, @ekizy,
I've created some tests for the generics functionality, but it's quite a big one at that, which means we will need to test it thoroughly and review the code to make sure there aren't some stupid mistakes or bugs.

If you'd like to speed up the process, you can help by using this branch `feature/generics` and trying out some unusual scenarios that you think of. If you create a pull request to the `feature/generics` branch with some additional tests of said scenarios, we would greatly appreciate that.

ekizy:
Thanks for the response @MatyasKriz ,

That sounds nice. I would gladly look in to your tests. Hope that process will be finished sooner. In my project I was tucked with the generic mocking too unfortunately.

MatyasKriz:
Until recently, this branch only supported class and protocol-wide generics, however, method generics along with support for where clause has been added and is ready for testing.

This is a pretty big change and generics is no small fish, so if you'd like to help us test it out a bit, we will greatly appreciate any feedback, especially found bugs.

pstasiak:
@MatyasKriz I'm testing your branch in one of my projects. When generating stubs for protocol defined as:
```swift
protocol Connecting {
    func assembleView() -> UIViewController
}
```
Cuckoo generates following stubing code:
```swift
 class ConnectingStub: Connecting {
     func assembleView() -> UIVie where roller  {
        return DefaultValueRegistry.defaultValue(for: UIVie.self)
    }   
}
```
There are some other method signatures mismatches like:
```swift
func assembleView() -> Cuckoo.ProtocolStubFunction<(), UIVie>
```

MatyasKriz:
Oops 😄 

This is what I'm talking about when I said it needs testing, thanks a lot for reporting this.

I had to create a return signature parser for proper generic methods support and didn't want to make it too complex to understand. Well, this is the result now. I will have to take a look at the parser and fix this.

I really appreciate you testing the generics branch, @pstasiak. I will try to fix it as soon as I can. I can't think of a workaround at the moment, so I will understand, if you decide to stop testing this branch.

If you don't mind not having the class-wide and protocol-wide generics, you can downgrade to the `feature/optionals` until this is fixed, however, if you don't mind, I'd greatly appreciate even more testing after this is fixed. Cheers!

pstasiak:
@MatyasKriz Thank You for the quick response :) Can You point out the object / function / file responsible for return signature parser?

MatyasKriz:
@pstasiak If I remember correctly, it should be located in `Tokenizer.swift` in generator's source code. Specifically in the giant `switch` when converting `Method` kind to Cuckoo's internal representation. The type to narrow the search down is `ReturnSignature`.

jamezilla:
This branch didn't work for me trying to generate a mock for this protocol:
```
import Networking
import RxSwift

public protocol GraphQLExecutorProvider {
    func executeCustomGraphQLQuery<Request: GraphQLQuery, Response: Decodable>(_ query: Request, api: GraphQLExecutor.GraphQLEndpoint) -> Single<Response>
}
```
This likely would have worked if the generic type was named better. The compiler error was: `'Request' is ambiguous for type lookup in this context`

I got it to work with some light modifications. I had to add explicit generics back to the function declarations. For example, I had to change from:
```
public func executeCustomGraphQLQuery(_ query: Request, api: GraphQLExecutor.GraphQLEndpoint)  -> Single<Response>
```
to this:
```
public func executeCustomGraphQLQuery<Request: GraphQLQuery, Response: Decodable>(_ query: Request, api: GraphQLExecutor.GraphQLEndpoint)  -> Single<Response>
```

And from this:
```
func executeCustomGraphQLQuery<M1: Cuckoo.Matchable, M2: Cuckoo.Matchable>(_ query: M1, api: M2) -> Cuckoo.ProtocolStubFunction<(Request, GraphQLExecutor.GraphQLEndpoint), Single<Response>> where M1.MatchedType == Request, M2.MatchedType == GraphQLExecutor.GraphQLEndpoint
```
to this:
```
func executeCustomGraphQLQuery<M1: Cuckoo.Matchable, M2: Cuckoo.Matchable, Request: GraphQLQuery, Response: Decodable>(_ query: M1, api: M2) -> Cuckoo.ProtocolStubFunction<(Request, GraphQLExecutor.GraphQLEndpoint), Single<Response>> where M1.MatchedType == Request, M2.MatchedType == GraphQLExecutor.GraphQLEndpoint
```

The changes above made this particular class work, but generated invalid mocks for many other classes which had RxSwift return types --- mocks which are generated correctly under 0.12.0. The errors were all the same: 
```
Generic type '__DoNotUse' specialized with too few type parameters (got 1, but expected 2)
```
and the generated code:
```
func authenticate<M1: Cuckoo.Matchable>(with userAuthenticationRequest: M1) -> Cuckoo.__DoNotUse<Single<DataResponse<UserAuthenticationResponse>>> where M1.MatchedType == UserAuthenticationRequest
```
And this is the protocol interface it was trying to generate:
```
func authenticate(with userAuthenticationRequest: UserAuthenticationRequest) -> Single<DataResponse<UserAuthenticationResponse>>
```

MatyasKriz:
@jamezilla Hey there, I'm not sure about the first error, but the second one seems to occur because of the upgrade to new Cuckoo while using old generator. At the moment we don't have a system in place to check whether cuckoo and generator versions match. This allows using different versions of Cuckoo framework and generator as long as they are compatible. I'm not sure, if anyone takes advantage of it, so I'm thinking of adding a version check to the script.
Anyway, to rebuild the generator using new source code, simply add `--clean` to the run script in your build phases. I recommend removing this after first run because it will increase build time.
Also, after adding function-scope generics, I introduced some bugs as well that @pstasiak found and I didn't get to fixing completely yet.

MatyasKriz:
I've addressed the bugs and this PR has been merged. Give it a try!

PR #257, opened 2019-02-22
Support spaces in script path in build_generator

dmazzoni:
The ```build_generator``` script did not support spaces in the script path, and failed at the ```pushd``` command.


MatyasKriz:
Merged and released. Thanks again!

PR #261, opened 2019-03-12
Add support for inout method parameters.

### LINKED ISSUE #260: Compilation error for "inout" arguments
Hey, I am using GRDB library for storing my models in database. And also I want to mock those models using Cuckoo.

The issue is that GRDB requires me to override the following function:

```
open func encode(to container: inout PersistenceContainer) {}
```

Cuckoo, on the other hand, tries to mock it and gives me compilation error in GeneratedMocks file. I guess it simply cannot handle `inout` yet.

Moreover, I cannot even exclude this function from mocking, because I don't really need mock for this function.

I also tried to mark this overridden function as `final` but Cuckoo still cannot handle this correctly.

MatyasKriz:
Closes #260.

### DIFF
--- a/Tests/Source/TestedClass.swift
+++ b/Tests/Source/TestedClass.swift
@@ -219,6 +219,20 @@ class FinalFields {
     final var field: Int? = nil
 
     final func function() {
-        
+
+    }
+}
+
+class InoutMethod {
+    func inoutko(param: inout Int) {
+
+    }
+
+    func inoutkoMultiple(param1: inout Int, param2: inout String, param3: Void) {
+
+    }
+
+    func inoutkoClosure(param: (inout Int) -> Void) {
+
     }
 }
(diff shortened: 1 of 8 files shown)


Tests/Source/TestedClass.swift
```diff
@@ -219,6 +219,20 @@ class FinalFields {
     final var field: Int? = nil
 
     final func function() {
-        
+
+    }
+}
+
+class InoutMethod {
+    func inoutko(param: inout Int) {
```
TadeasKriz: `inoutko` 😄

MatyasKriz:
Thanks for the review. By the way, I have a feeling that Travis is compiling the project in Swift 3, that's why it's failing. Maybe we should mark next releases as Swift 4 and up, making 0.12.1 the last Swift 3 version.

TadeasKriz:
Sounds good.

PR #271, opened 2019-04-03
Add property attributes to stubs and proxies

dmazzoni:
Adds property attributes to the templates of stubbing proxies, verification proxies and stubs.

For example, this is required for the following example protocol to be mockable with a deployment target lower than the iOS version specified in the `@available` attribute (suppose that `Baz` was introduced with iOS 12):

```
protocol Bar {
    
    @available(iOS 12.0, *)
    var foo: Baz? { get }
}
```

If the `@available` attribute is not included in stubs and proxies, Cuckoo's output file will not compile.

### DIFF
--- a/Generator/Source/CuckooGeneratorFramework/Templates/NopImplStubTemplate.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Templates/NopImplStubTemplate.swift
@@ -9,6 +9,9 @@ extension Templates {
     static let noImplStub = """
 {{container.accessibility}} class {{ container.name }}Stub: {{ container.name }} {
     {% for property in container.properties %}
+    {% for attribute in property.attributes %}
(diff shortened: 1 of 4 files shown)


MatyasKriz:
Huh, nice catch. I don't know why I didn't think to include that in the tests. Do you think you could also create an annotated variable inside the `TestedClass` in `TestedClass.swift` as a sort of fail-safe that tests this scenario?

TadeasKriz:
This looks very good! Could you please add tests @MatyasKriz mentioned? Thanks you for the contribution!

dmazzoni:
No problem, I'll update the PR tomorrow!

TadeasKriz:
iOS 42 :D

MatyasKriz:
Merged, thanks!

Issue #264, opened 2019-03-19
Building the generator fails with the Xcode 10.2 toolchain


dmazzoni:
It seems that building the generator always fails when using the Xcode 10.2 toolchain.

This is the relevant part of Cuckoo's log:

```
Fetching https://github.com/kylef/PathKit.git
Fetching https://github.com/Quick/Nimble.git
Fetching https://github.com/Carthage/Commandant.git
Fetching https://github.com/jpsim/SourceKitten.git
Fetching https://github.com/Quick/Quick.git
Fetching https://github.com/drmohundro/SWXMLHash.git
Fetching https://github.com/kylef/Stencil.git
Fetching https://github.com/jpsim/Yams.git
Fetching https://github.com/TadeasKriz/FileKit.git
Fetching https://github.com/kylef/Spectre.git
Fetching https://github.com/antitypical/Result.git
https://github.com/TadeasKriz/FileKit.git @ develop: error: manifest parse error(s):
/tmp/TemporaryFile.acKJOQ.swift:32:14: error: extra argument 'exclude' in call
    exclude: [
             ^
```

I suppose that the error is due to the fact that FileKit is written in Swift 3, which is not supported by the Xcode 10.2 compiler. I did not open a PR with the Swift 4 migration because FileKit's upstream repository (`nvzqz/FileKit`) has already been migrated to Swift 4.2.

MatyasKriz:
Hi, thanks for checking, now it won't just stop working on us suddenly after using new Swift.
I investigated why we were using our own FileKit and only saw commits related to compilability in Xcode 9, which I'm sure is working in their master as of the moment. We discussed it and decided to switch back to the original FileKit.
Since you found this out, would you like to create a PR updating FileKit to point at the original repo again?

dmazzoni:
At this moment it would only fix part of the issue. The original FileKit repo has already been migrated to Swift 4.2, but its `Package.swift` manifest file uses the v3 format, which seems to be incompatible with the Xcode 10.2 toolchain. In fact, the actual reason of the error I reported when opening this issue is the manifest file format, and not the project's Swift version.

Yesterday I forked FileKit and updated its `Package.swift` file so as to make it compatible with Xcode 10.0 to 10.2; before opening a PR on FileKit I'll also try to ensure compatibility with Xcode 9.4.

dmazzoni:
I've just opened the PR: nvzqz/FileKit#68.

dmazzoni:
[...]

PR #272, opened 2019-04-08
Depend on original FileKit repo to build Generator

dmazzoni:
The manifest file of the original FileKit repository has been updated to the 4.2 syntax. This PR updates the Generator's `Package.swift` file to point to the original FileKit repository, fixing the error reported in #264.

FileKit's updated manifest file is compatible with the Xcode toolchain from version 10.0 to 10.2.

### DIFF
--- a/Generator/Package.resolved
+++ b/Generator/Package.resolved
@@ -12,10 +12,10 @@
       },
       {
         "package": "FileKit",
-        "repositoryURL": "https://github.com/TadeasKriz/FileKit.git",
+        "repositoryURL": "https://github.com/nvzqz/FileKit.git",
         "state": {
           "branch": "develop",
-          "revision": "0acc6e7c336bbd2336c2f9670564dac9ec9c03fe",
+          "revision": "48b5ddb287f131a5c628badc819b880453f94449",
           "version": null
         }
       },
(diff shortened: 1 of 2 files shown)


jamezilla:
This seems to be failing CI because it's trying to build swift 4.2 code with the swift 4.0 toolchain.

Otherwise, this PR looks ready to go. @dmazzoni's PR into FileKit was merged on 3/20: https://github.com/nvzqz/FileKit/pull/68

jamezilla:
This branch does not currently compile. I had to add `Result` as an explicit dependency to Package.swift file to get this to compile:
https://github.com/jamezilla/Cuckoo/commit/e3da7ab56c7166cdb9995c86704d242ca8c6d716#diff-3bdc43e1bd246d5e6442dbc1386eab9dR14

I presume some dependency's minor version bump caused `Result` to be dropped from the Package.resolve graph.

dmazzoni:
@jamezilla It should be unrelated to the FileKit reference change, maybe it's due to an update in Commandant?

TadeasKriz:
We should use Result from Swift 5.

jamezilla:
@dmazzoni Yes, it's unrelated to the FileKit change.

@TadeasKriz I think moving to the native result type is the right plan. If I understand correctly, as long as you're on the 10.2 toolchain, Swift 5 features [mostly work](https://docs.swift.org/swift-book/GuidedTour/Compatibility.html) with consuming projects written in Swift 4.x. It should have a minimal impact.

MatyasKriz:
Result should have been removed from the README in the section of used libraries.

PR #273, opened 2019-04-30
fix curl malformed url error when no github token is passed

jamezilla:
When `REQUEST_HEADER` is null, it gets expanded to `''` (double single quotes) which causes curl to throw an error: `curl: (3) <url> malformed`.

On my macOS Mojave machine this causes a lot of problems. The second invocation of curl in the `run` script outputs the whole binary to stderr and dumps all that garbage to my screen.

This patch does the variable expansion "only if the variable is defined" and is [supported by all bash shells](http://mywiki.wooledge.org/BashFAQ/073).

### DIFF
--- a/run
+++ b/run
@@ -12,9 +12,9 @@ function perform_curl {
   fi
   echo "$1"
 
-  DOWNLOAD_URL=$(curl "$REQUEST_HEADER" "$1" | grep -oe '"browser_download_url":\s*"[^" ]*"' | grep -oe 'http[^" ]*' | grep "$FILE_NAME" | head -1)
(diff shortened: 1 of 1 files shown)


MatyasKriz:
Nice one! I somehow didn't trigger this when testing the PR that added this functionality. I'll make sure to include your fix in the next release.

MatyasKriz:
Update on this, I didn't properly test the previous PR and I'm not sure the auth token was actually used back then. I couldn't use your solution because it added some unnecessary apostrophes and when testing with real tokens, GitHub didn't report that I had used them.

The solution I went with was using bash array that is just slapped into the command calls and it looks just like it should. It will be pushed later today.

Thank you for the PR anyway, please don't let this deter you in any way from further contributing to Cuckoo.

PR #278, opened 2019-05-13
Change build_generator to work with Swift 5

henriquegouveia:
The current `build_generator` script which executes `swift build` command to generator the `cuckoo_generator` doesn't work with Swift 5 so the new command line `swift build --configuration release --static-swift-stdlib` is enough to build the `cuckoo_generator` and also keep it distributable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,12 @@
 language: objective-c
-osx_image: xcode9
+osx_image: xcode10.2
 xcode_project: Cuckoo.xcodeproj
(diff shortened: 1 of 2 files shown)


MatyasKriz:
Hey there. Great job! It didn't occur to me that this could become a problem going forward.

We checked compatibility with Swift 4.2 and it works flawlessly, so I'll merge this into the next release.

Thanks for your PR!

MatyasKriz:
Merged, thanks!

PR #282, opened 2019-05-16
Add tests for stubbing optional read only properties.

adotor:
Tests for stubbing a readOnly-Property that is optional.

The nil-case fails compiling while the value-case fails with a missing stub.


MatyasKriz:
Hey, thanks a bunch for this PR! Now it makes sense and will hopefully prevent us from breaking this again in the future. I have fixed it, the build is successful, and the tests are passing.

This fix is released as we speak in the `1.0.3` version.

Thanks again for the issue report and tests that highlight this bug.

PR #315, opened 2019-10-16
Switch swift_version to 5.0 and add LIBRARY_SEARCH_PATHS in xcconfig

artyom-razinov:
This fixes running tests on iOS 10/11/12 with Cuckoo when tests are built using Xcode 11.

It is a minimal required fix. I checked this, nothing can be removed from these changes to fix the issue.

I think it is okay to switch to Swift 5 at the moment. It is the biggest change in terms of compatibility, it has stable ABI, etc. I was trying to keep my project in Swift 5, but I gave up after several days. You can make a tag named swift-4.2 for current HEAD in Cuckoo. People who use your project will still be able use your project.

I think there is no point in holding on to the past now.

PR #317, opened 2019-10-24
Fix closure generation, return type is required in `withoutActuallyEs…

Pubbus:
…caping` from swift 5.1 (XCode11) Brightify#313

### DIFF
--- a/Generator/Source/CuckooGeneratorFramework/Generator.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Generator.swift
@@ -49,8 +49,11 @@ public struct Generator {
 
         ext.registerFilter("openNestedClosure") { (value: Any?, arguments: [Any?]) in
             guard let parameters = value as? [MethodParameter] else { return value }
-
-            let s = self.openNestedClosure(for: parameters, throwing: arguments.first as? Bool)
+            var returnType: String?
+            if arguments.count > 1 {
+              returnType = arguments[1] as? String
+            }
+            let s = self.openNestedClosure(for: parameters, throwing: arguments.first as? Bool, returnType: returnType)
             return s
         }
 
@@ -108,13 +111,15 @@ public struct Generator {
         return type.replacingOccurrences(of: "!", with: "?")
     }
 
-    private func openNestedClosure(for parameters: [MethodParameter], throwing: Bool? = false) -> String {
+    private func openNestedClosure(for parameters: [MethodParameter], throwing: Bool? = false, returnType: String?) -> String {
         var fullString = ""
         for (index, parameter) in parameters.enumerated() {
             if parameter.isClosure && !parameter.isEscaping {
                 let indents = String(repeating: "\t", count: index)
                 let tries = (throwing ?? false) ? " try " : " "
-                fullString += "\(indents)return\(tries)withoutActuallyEscaping(\(parameter.name), do: { (\(parameter.name): @escaping \(parameter.type)) in\n"
+                let returns = (returnType != nil) ? " -> \(returnType!)" : ""
+
+                fullString += "\(indents)return\(tries)withoutActuallyEscaping(\(parameter.name), do: { (\(parameter.name): @escaping \(parameter.type))\(returns) in\n"
             }
         }
         return fullString
--- a/Generator/Source/CuckooGeneratorFramework/Generator.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Generator.swift
@@ -49,12 +49,11 @@ public struct Generator {
 
         ext.registerFilter("openNestedClosure") { (value: Any?, arguments: [Any?]) in
             guard let parameters = value as? [MethodParameter] else { return value }
-            var returnType: String?
-            if arguments.count > 1 {
-              returnType = arguments[1] as? String
+            var returnType: WrappableType = WrappableType(parsing: "")
+            if arguments.count > 1, let unwrappedWrapableType = arguments[1] as? WrappableType {
+              returnType = unwrappedWrapableType
             }
-            let s = self.openNestedClosure(for: parameters, throwing: arguments.first as? Bool, returnType: returnType)
-            return s
+            return self.openNestedClosure(for: parameters, throwing: arguments.first as? Bool, returnType: returnType)
         }
 
         ext.registerFilter("closeNestedClosure") { (value: Any?) in
@@ -111,15 +110,18 @@ public struct Generator {
         return type.replacingOccurrences(of: "!", with: "?")
     }
 
-    private func openNestedClosure(for parameters: [MethodParameter], throwing: Bool? = false, returnType: String?) -> String {
+    private func openNestedClosure(for parameters: [MethodParameter], throwing: Bool? = false, returnType: WrappableType) -> String {
         var fullString = ""
         for (index, parameter) in parameters.enumerated() {
             if parameter.isClosure && !parameter.isEscaping {
                 let indents = String(repeating: "\t", count: index)
                 let tries = (throwing ?? false) ? " try " : " "
-                let returns = (returnType != nil) ? " -> \(returnType!)" : ""
+                var returnSignature = returnType.sugarized
+                if !returnSignature.isEmpty {
+                  returnSignature = " -> \(returnSignature)"
+                }
 
-                fullString += "\(indents)return\(tries)withoutActuallyEscaping(\(parameter.name), do: { (\(parameter.name): @escaping \(parameter.type))\(returns) in\n"
+                fullString += "\(indents)return\(tries)withoutActuallyEscaping(\(parameter.name), do: { (\(parameter.name): @escaping \(parameter.type))\(returnSignature) in\n"
             }
         }
         return fullString
--- a/Generator/Source/CuckooGeneratorFramework/Templates/MockTemplate.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Templates/MockTemplate.swift
@@ -103,7 +103,7 @@ extension Templates {
     {{ attribute.text }}
     {% endfor %}
     {{ method.accessibility }}{% if container.isImplementation and method.isOverriding %} override{% endif %} func {{ method.name }}{{ method.genericParameters }}({{ method.parameterSignature }}) {{ method.returnSignature }} {
-        {{ method.parameters|openNestedClosure:method.isThrowing }}
+        {{ method.parameters|openNestedClosure:method.isThrowing, :method.returnType }}
     return{% if method.isThrowing %} try{% endif %} cuckoo_manager.call{% if method.isThrowing %}{{ method.throwType|capitalize }}{% endif %}("{{method.fullyQualifiedName}}",
             parameters: ({{method.parameterNames}}),
             escapingParameters: ({{method.escapingParameterNames}}),
--- a/Generator/Source/CuckooGeneratorFramework/Tokens/Method.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Tokens/Method.swift
@@ -105,8 +105,7 @@ public extension Method {
                     if !returnSignature.isEmpty {
                       if returnSignature != "Void" {
                         returnSignature = " -> " + returnSignature
-                      }
-                      else {
+                      } else {
                         returnSignature = ""
                       }
                     }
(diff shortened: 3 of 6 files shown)


MatyasKriz:
Hi, thanks for the PR, I am grateful on behalf of the Cuckoo community.

I'll have to deal with the Carthage situation and then merge in PR #315 until I can get back to you. That'll hopefully be during today.

Pubbus:
That sounds great. Have a great day.

Generator/Source/CuckooGeneratorFramework/Tokens/Method.swift
```diff
@@ -98,7 +98,13 @@ public extension Method {
             if parameter.isClosure && !parameter.isEscaping {
                 let parameterCount = parameter.closureParamCount
                 let parameterSignature = parameterCount > 0 ? (1...parameterCount).map { _ in "_" }.joined(separator: ", ") : "()"
-                return "{ \(parameterSignature) in fatalError(\"This is a stub! It's not supposed to be called!\") }"
+                var returns: String?
+
+                if !parameter.type.sugarized.isEmpty {
+                  returns = parameter.type.sugarized.components(separatedBy: "->").last
+                }
```
MatyasKriz: In conjunction with the ternary operator down below, we can minimize this into a single named constant (e.g. `returnSignature`). It should look something like this:
```
let returnType = parameter.type.sugarized
let returnSignature: String
if returnType.isEmpty || returnType == "Void" {
    returnSignature = ""
} else {
    returnSignature = " -> \(returnType)"
}
```

Generator/Source/CuckooGeneratorFramework/Tokens/Method.swift
```diff
@@ -98,7 +98,13 @@ public extension Method {
             if parameter.isClosure && !parameter.isEscaping {
                 let parameterCount = parameter.closureParamCount
                 let parameterSignature = parameterCount > 0 ? (1...parameterCount).map { _ in "_" }.joined(separator: ", ") : "()"
-                return "{ \(parameterSignature) in fatalError(\"This is a stub! It's not supposed to be called!\") }"
+                var returns: String?
+
+                if !parameter.type.sugarized.isEmpty {
+                  returns = parameter.type.sugarized.components(separatedBy: "->").last
```
MatyasKriz: I'm not really sure if this will play nice with returning closures. Have you tested return functions with signatures `f(p:) -> () -> Void` or even `f(p:) -> (() -> Void)?`?

I have a feeling that we'll need to add support for parsing closures for this to work properly.

Generator/Source/CuckooGeneratorFramework/Tokens/Method.swift
```diff
@@ -98,7 +98,13 @@ public extension Method {
...
+                }
+
+                return "{ \(parameterSignature) \(returns != nil && returns != "Void" ? " -> \(returns!) " : "") in fatalError(\"This is a stub! It's not supposed to be called!\") }"
```
MatyasKriz: Expanding on the comment above, we need to eliminate the ternary operator here as well as the unwanted duplicate space. If you remove the space between `\(parameterSignature)` and the following string interpolation, all will be well as we've added the space to the variable already.

Generator/Source/CuckooGeneratorFramework/Generator.swift
```diff
@@ -49,8 +49,11 @@ public struct Generator {
...
         ext.registerFilter("openNestedClosure") { (value: Any?, arguments: [Any?]) in
             guard let parameters = value as? [MethodParameter] else { return value }
-
-            let s = self.openNestedClosure(for: parameters, throwing: arguments.first as? Bool)
+            var returnType: String?
+            if arguments.count > 1 {
+              returnType = arguments[1] as? String
+            }
+            let s = self.openNestedClosure(for: parameters, throwing: arguments.first as? Bool, returnType: returnType)
             return s
```
MatyasKriz: I know this isn't part of this PR, but this has escaped my attention. Please instead of using a named constant, return the result immediately. Thank you.

Generator/Source/CuckooGeneratorFramework/Generator.swift
```diff
@@ -49,8 +49,11 @@ public struct Generator {
 
         ext.registerFilter("openNestedClosure") { (value: Any?, arguments: [Any?]) in
```
MatyasKriz: I've used the search functionality on GitHub and it seems that this filter is used in one place and one place only.

I would propose rather than to add more arguments that are taken from the method anyway, we should pass the whole `Method` instead.

If you have any questions about what I mean, don't hesitate to ask.

Generator/Source/CuckooGeneratorFramework/Templates/MockTemplate.swift
```diff
@@ -103,7 +103,7 @@ extension Templates {
     {{ attribute.text }}
     {% endfor %}
     {{ method.accessibility }}{% if container.isImplementation and method.isOverriding %} override{% endif %} func {{ method.name }}{{ method.genericParameters }}({{ method.parameterSignature }}) {{ method.returnSignature }} {
-        {{ method.parameters|openNestedClosure:method.isThrowing }}
+        {{ method.parameters|openNestedClosure:method.isThrowing, :method.returnType }}
```
MatyasKriz: Restructuring the `openNestedClosure` filter to take in just the method, this call should then look like `{{ method|openNestedClosure }}`.

Generator/Source/CuckooGeneratorFramework/Generator.swift
```diff
@@ -108,13 +111,15 @@ public struct Generator {
...
 
-    private func openNestedClosure(for parameters: [MethodParameter], throwing: Bool? = false) -> String {
+    private func openNestedClosure(for parameters: [MethodParameter], throwing: Bool? = false, returnType: String?) -> String {
         var fullString = ""
         for (index, parameter) in parameters.enumerated() {
             if parameter.isClosure && !parameter.isEscaping {
                 let indents = String(repeating: "\t", count: index)
                 let tries = (throwing ?? false) ? " try " : " "
-                fullString += "\(indents)return\(tries)withoutActuallyEscaping(\(parameter.name), do: { (\(parameter.name): @escaping \(parameter.type)) in\n"
+                let returns = (returnType != nil) ? " -> \(returnType!)" : ""
```
MatyasKriz: Please avoid implicitly unwrapping optionals. I know that right now it undoubtedly has value, but during another restructuring, someone might break the condition and we might not catch it.

The way we do this simple optional unwrapping into a constant is like so:
```
let returnSignature = returnType.map { " -> \($0)" } ?? ""
```

Generator/Source/CuckooGeneratorFramework/Tokens/Method.swift
Pubbus: Agreed.

Pubbus:
Thank for reviewing, I'm about to check and do some update on this, will let you know after that.

Generator/Source/CuckooGeneratorFramework/Generator.swift
Pubbus: We should not use `map` here, map extension will loop over characters and insert the closure result into every element then return an array of elements. I have short it out on new commit.

Generator/Source/CuckooGeneratorFramework/Generator.swift
Pubbus: Updated

Generator/Source/CuckooGeneratorFramework/Tokens/Method.swift
Pubbus: Updated

Generator/Source/CuckooGeneratorFramework/Tokens/Method.swift
Pubbus: Addressed on new commit. I've added a playground to test against that.

```
let test1 = "(() -> Int) -> Int"
print(extractClosureReturnType(parameter: test1) == "Int")
let test2 = "f(p:) -> () -> Void"
print(extractClosureReturnType(parameter: test2) == "() -> Void")
let test3 = "f(p:) -> (() -> Void)?"
print(extractClosureReturnType(parameter: test3) == "(() -> Void)?")
let test4 = "(() -> Int) -> () -> Void"
print(extractClosureReturnType(parameter: test4) == "() -> Void")
```

Generator/Source/CuckooGeneratorFramework/Generator.swift
Pubbus: I've tried with `{{ method|openNestedClosure }}` `ext.registerFilter("openNestedClosure") { (value: Any?)` `value is Method` in there but got no luck, lot of errors occurred. Can you try that after this MR?

Generator/Source/CuckooGeneratorFramework/Templates/MockTemplate.swift
Pubbus: Explained above

Pubbus:
Hey @MatyasKriz, I've updated the MR. Can you take a look at this?

Generator/Source/CuckooGeneratorFramework/Generator.swift
MatyasKriz: `returnType` is an optional, you are mapping on its value and leaving `nil` otherwise. You mustn't put the question mark before the map for this to work correctly.

Generator/Source/CuckooGeneratorFramework/Tokens/Method.swift
MatyasKriz: Why have you added a playground instead of adding this case to the tests instead?

Generator/Source/CuckooGeneratorFramework/Generator.swift
Pubbus: I've refactored in the newest commit.

Generator/Source/CuckooGeneratorFramework/Tokens/Method.swift
Pubbus: Make sense, I'm about to remove this playground and add test cases instead.

Pubbus: Replaced. Please check.

MatyasKriz:
Hello, @Pubbus. Do not worry, for I haven't forgotten about Cuckoo just yet. 🙂

I've been sick for 3 days now, so I didn't have the time to review your changes yet. Depending on my state today I might get to it, please be patient with me, thank you.

Pubbus:
Take your time and get well soon. I didn't mean to rush you. 😄

Generator/Source/CuckooGeneratorFramework/Tokens/Method.swift
```diff
@@ -134,4 +147,22 @@ public extension Method {
...
+    var closeBracketCount = 0
+    for i : Int in 0..<parameter.count {
+      let index = parameter.index(parameter.startIndex, offsetBy: i)
+      if parameter[index] == "(" {
+        openBracketCount += 1
+      }
+      else if (parameter[index] == ")") {
+        closeBracketCount += 1
+        if openBracketCount == closeBracketCount {
+         let afterClosure = String(parameter[parameter.index(after: index)..<parameter.endIndex])
```
MatyasKriz: Please keep 4-space indentation consistently throughout your changes as to stay true with the rest of the codebase. Thank you.

Generator/Source/CuckooGeneratorFramework/Utils.swift
```diff
@@ -18,6 +18,17 @@ extension String {
         return components(separatedBy: occurence).first
     }
 
+    func matches(for regex: String) -> [String] {
+      do {
```
MatyasKriz: You may omit the `do catch`, because as far as I know, `NSRegularExpression.init` throws only when the regex is invalid, in which case I think it's better to throw as well instead of silently returning "no matches" array.

Tests/ClassTest.swift
```diff
@@ -153,10 +153,16 @@ class ClassTest: XCTestCase {
     func testWithClosure() {
         stub(mock) { mock in
             when(mock.withClosure(anyClosure())).then { $0("a") }
+            when(mock.withClosureReturnAVoidClosure(anyClosure())).then { $0("a")()}
+            when(mock.withClosureReturnAnIntClosure(anyClosure())).then { $0("a")(1)}
         }
 
         XCTAssertEqual(mock.withClosure { _ in 1 }, 1)
+        XCTAssertEqual(mock.withClosureReturnAVoidClosure {_ in { return 1 }}, 1)
+        XCTAssertEqual(mock.withClosureReturnAnIntClosure { _ in { _ in return 1} }, 1)
```
MatyasKriz: Please separate the braces with spaces from apart each other.

Generator/Source/CuckooGeneratorFramework/Tokens/Method.swift
```diff
@@ -134,4 +147,22 @@ public extension Method {
             "genericParameters": isGeneric ? "<\(genericParametersString)>" : "",
         ]
     }
+  private func extractClosureReturnType(parameter: String) -> String? {
+    var openBracketCount = 0
+    var closeBracketCount = 0
```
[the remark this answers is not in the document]
Pubbus: It is great! I about to update it now.

Pubbus:
Thanks for reviewing 😁, I have addressed all of them.

Pubbus:
Thank you!!!

MatyasKriz:
I checked the code and one thing that stands out is that the errors depend on the `escapingParameters` part where you've added the return type of the closure. Even so I've kept the `withoutActuallyEscaping` explicit return type, as it doesn't hurt and might save a bit of time for large generated files.

In the future I'd like to move the closure return type parsing out of `Method.swift` where it doesn't really belong. Tokenizer should do it instead and pass the information in some data structure (similar to `WrappableType`).

However, at the moment I don't have enough time to get to it and get it over with. It'll have to wait for now as there are more pressing matters that need my attention.

I've also added some ludicrous testing code that uses closures that hopefully won't ever be used, but just in case.

Thanks again for your pull request! I appreciate the effort you put into this and am grateful that you support the Cuckoo community.

PR #319, opened 2019-11-08
Fix typo regarding things overridable by Cuckoo

schnaser:
I think this is what was meant? The double negative threw me

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ We are still missing support for some important features like:
 - [x] Objective-C support utilizing OCMock
 
 ## What will not be supported
-Due to the limitations mentioned above, none of the things that can't be overridden can't be supported. This includes:
+Due to the limitations mentioned above, none of the things that can't be overridden are supportable by Cuckoo. This includes:
 - `struct` - workaround is to use a common protocol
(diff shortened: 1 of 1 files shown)


MatyasKriz:
Ah, I see, triple negative no less, I would've been confused as well.

README.md
```diff
@@ -32,7 +32,7 @@ We are still missing support for some important features like:
 - [x] Objective-C support utilizing OCMock
 
 ## What will not be supported
-Due to the limitations mentioned above, none of the things that can't be overridden can't be supported. This includes:
+Due to the limitations mentioned above, none of the things that can't be overridden are supportable by Cuckoo. This includes:
```
MatyasKriz: This is better, but I'd like to replace this with a more structurally sound sentence that might be completely different from this one. Might even be better to rid us of the negatives altogether.

If you can think of something, don't hesitate to update the PR and ping me. I'll try to think about this as well and will let you know.

MatyasKriz:
@schnaser thank you so much for this PR. Pointing out the overly complicated sentence prompted me to restructure the whole README a bit but most importantly rewrite this sentence to make it hopefully easy to comprehend.

Issue #321, opened 2019-12-05
Prefix mock- and stubnames


emielv:
Hi all,

My team wants to slowly integrate the Cuckoo framework into our project for generating mocks. We want to support tests that still use our own mocks and new / refactored tests that use Cuckoo mocks. The problem we are facing is that there are conflicts between mocks we wrote ourselves and mocks that Cuckoo generates e.g. same classname for the mocks and it also reduces readability because it’s not immediately obvious if you are working with Cuckoo mocks or not. 
Is it possible / is there a way to add a prefix to the classnames of the generated mocks e.g. generating a mock for protocol Abc will generate CuckooMockAbc / CKMockAbc?

Thanks in advance!

MatyasKriz:
Hey, that's not a bad idea. I'd welcome a PR that would implement this to make one's life easier.

You need to start by running `make dev` in project root, which creates generator project file for the Cuckoo workspace.

I think you'd have to edit `GenerateMocksCommand.swift` and define a command named "mock-prefix" perhaps that would register the prefix given by user, optional all the way to the mocks generation and then either using that or the default "Mock".

Let me know if you get stuck or need some clarification, I'll be happy to help.

emielv:
Hello @MatyasKriz ,
I created a PR regarding this feature and would love some feedback.

Thanks!

PR #322, opened 2019-12-15
Added option to prefix classnames of mocks and stubs

emielv:
Related to https://github.com/Brightify/Cuckoo/issues/321

### DIFF
--- a/Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
+++ b/Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
@@ -13,12 +13,12 @@ import FileKit
 import CuckooGeneratorFramework
 import Foundation
 
-private func curry<P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11, P12, R>
-    (_ f: @escaping (P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11, P12) -> R)
-    -> (P1) -> (P2) -> (P3) -> (P4) -> (P5) -> (P6) -> (P7) -> (P8) -> (P9) -> (P10) -> (P11) -> (P12) -> R {
-        return { p1 in { p2 in { p3 in { p4 in { p5 in { p6 in { p7 in { p8 in { p9 in { p10 in { p11 in { p12 in
-            f(p1, p2, p3, p4, p5, p6, p7, p8, p9, p10, p11, p12)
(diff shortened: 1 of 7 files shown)


Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
```diff
@@ -13,12 +13,12 @@ import FileKit
...
-    -> (P1) -> (P2) -> (P3) -> (P4) -> (P5) -> (P6) -> (P7) -> (P8) -> (P9) -> (P10) -> (P11) -> (P12) -> R {
-        return { p1 in { p2 in { p3 in { p4 in { p5 in { p6 in { p7 in { p8 in { p9 in { p10 in { p11 in { p12 in
-            f(p1, p2, p3, p4, p5, p6, p7, p8, p9, p10, p11, p12)
-        } } } } } } } } } } } }
+private func curry<P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11, P12, P13, R>
+    (_ f: @escaping (P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11, P12, P13) -> R)
+    -> (P1) -> (P2) -> (P3) -> (P4) -> (P5) -> (P6) -> (P7) -> (P8) -> (P9) -> (P10) -> (P11) -> (P12) -> (P13) -> R {
+        return { p1 in { p2 in { p3 in { p4 in { p5 in { p6 in { p7 in { p8 in { p9 in { p10 in { p11 in { p12 in { p13 in
+            f(p1, p2, p3, p4, p5, p6, p7, p8, p9, p10, p11, p12, p13)
+            } } } } } } } } } } } } }
```
MatyasKriz: Please remove the 4 spaces in front of the braces.

MatyasKriz:
I've checked the code and I think it might be better (at least for now when the prefix is identical in all places) to not pass the prefix through the tokenizer as it's not really its job to know about it at all, same with the individual class and protocol declarations.

My proposal is passing the prefix as an optional parameter to the `generate` method on the `Generator` that is called in `GenerateMocksCommand.swift` and then passing that to the Stencil context, so that when prefix is passed, it's used as `{{ prefix }}`, nil-coalescing with empty string inside the Generator itself.

What do you think about it?

MatyasKriz:
Closed due to inactivity.

PR #323, opened 2019-12-16
Adds tvOS target to be used with Carthage

rodrigoff:
I've been trying to use Cuckoo on a tvOS project with Carthage, and noticed there wasn't a tvOS target, so I created them based on their iOS counterparts, changing only a few settings to their tvOS specifics, such as the base SDK and supported platforms.


MatyasKriz:
@rodrigoff thanks for the PR! I've merged and released it in `1.3.1`.

PR #332, opened 2020-02-03
Fail immediately if any command fails

kainjow:
I was having issues where the `swift build` command was failing but the script itself wouldn't fail. Using `set -e` will cause the shell script to immediately stop and exit if any command in the script itself fails.

### DIFF
--- a/build_generator
+++ b/build_generator
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
+
 # getting the script directory: https://stackoverflow.com/a/246128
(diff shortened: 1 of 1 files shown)


TadeasKriz:
Thank you for your PR. In the description you mention `set -x`, but the change contains `set -e`. Which one is correct? Thank you!

kainjow:
Oops, fixed!

MatyasKriz:
Thanks @kainjow for the PR! I've merged it and released it in `1.3.1`.

PR #336, opened 2020-03-06
Support Sequence Type result OCMock extension

nisin:
Support for like that code.

```
let locMock = objcStub(for: CLLocationManager.self) { stubber, mock in
            stubber.when(mock.monitoredRegions).thenReturn(Set<CLRegion>())
        }
```

### DIFF
--- a/Tests/OCMock/ObjectiveClassTest.swift
+++ b/Tests/OCMock/ObjectiveClassTest.swift
@@ -20,18 +20,20 @@ class ObjectiveClassTest: XCTestCase {
     }
 
     func testThenReturn() {
+        let subViews = [UIView()]
         let mock = objcStub(for: UIView.self) { stubber, mock in
             stubber.when(mock.endEditing(true)).thenReturn(true)
             stubber.when(mock.endEditing(false)).thenReturn(false)
-            stubber.when(mock.subviews).thenReturn([UIView]())
+            stubber.when(mock.subviews).thenReturn(subViews)
         }
 
         XCTAssertTrue(mock.endEditing(true))
         XCTAssertFalse(mock.endEditing(false))
-        XCTAssertTrue(mock.subviews.isEmpty)
+        XCTAssertEqual(mock.subviews,subViews)
         
         objcVerify(mock.endEditing(true))
         objcVerify(mock.endEditing(false))
+        objcVerify(mock.subviews)
     }
 
     func testThen() {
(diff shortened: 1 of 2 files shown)


MatyasKriz:
Hey, @nisin. Thanks for your contribution to Cuckoo! I'm thrilled that ObjC mocking from Swift is used in the community.
Would you mind adding a test case to the Cuckoo+OCMock target that would seal the deal and would let us verify each time that the functionality works as it should?
Thank you!

nisin:
Thank you for your comment.
I impliment the test.
I hope this is what you want.

Tests/OCMock/ObjectiveClassTest.swift
```diff
@@ -23,11 +23,13 @@ class ObjectiveClassTest: XCTestCase {
...
         }
 
         XCTAssertTrue(mock.endEditing(true))
         XCTAssertFalse(mock.endEditing(false))
-
+        XCTAssertTrue(mock.subviews.isEmpty)
+        
         objcVerify(mock.endEditing(true))
         objcVerify(mock.endEditing(false))
     }
```
MatyasKriz: It's been a while since I added these tests, so have patience please. Did you omit the objcVerify because it doesn't work or has there been other reason?

Tests/OCMock/ObjectiveClassTest.swift
```diff
@@ -23,11 +23,13 @@ class ObjectiveClassTest: XCTestCase {
         let mock = objcStub(for: UIView.self) { stubber, mock in
             stubber.when(mock.endEditing(true)).thenReturn(true)
             stubber.when(mock.endEditing(false)).thenReturn(false)
+            stubber.when(mock.subviews).thenReturn([UIView]())
```
MatyasKriz: If this works with just `[]`, use that, please.

nisin: Reconsidered and compared with the array containing the view.

Tests/OCMock/ObjectiveClassTest.swift
nisin: > objcVerify because it doesn't work or has there been other reason?
have No reason. I just overlooked it.

MatyasKriz:
Alright, this looks good to me. 🙂 I'll merge it when I get the chance.

MatyasKriz:
Alright, I added one more test for the `then` that `return`s in a closure and merged it into master. It doesn't warrant its own release at the moment, so we'll keep it in master until there are more changes (for example after fixing #322 and merging it).
Anyway, thanks a lot for your contribution, @nisin!

PR #338, opened 2020-03-21
Integrate `tuist`.

MatyasKriz:
Both iOS and macOS targets were successfully added. However, [tuist](https://github.com/tuist/tuist) does not support tvOS (or watchOS for that matter) for generating the targets/schemes.

I'm not sure how to proceed through that without some mad hacks, so either we wait with merging until **tuist** supports tvOS or blindly believe that the tests pass (frankly, I can't think of a change that might break tests just on one platform).

Either way, this was my first tuist integration, so I'd appreciate some feedback.


TadeasKriz:
Do you know how much time would it take to add tvOS and watchOS support for tuist via a PR?

MatyasKriz:
I've skimmed through the code and it appears that there was an attempt to add the support but it was killed (dead code), so I'll head on over there and ask them what the issue was.

TadeasKriz:
Thank you, much appreciated! Please glance through this file: 
https://github.com/Brightify/Hyperdrive/blob/feature/experiment/hyperdrive/Platform/Project.swift

I'm not sure if it generates tvOS target correctly, but could it be a solution for Cuckoo if this works?

Lastly, don't worry about watchOS, Cuckoo never supported it AFAIK. Since we require XCTest package which isn't available on watchOS, users are out of luck (unless Apple released it already).

However there's this project:
https://github.com/dfed/XCTest-watchOS

Might be an interesting one.

MatyasKriz:
Alright, it turns out that **tuist** does support tvOS as a platform, just not the deployment target, so that will need to be set manually before release.
However, this makes the tvOS target usable and allows for testing on other computers than where it's manually created. Good progress, thanks for the tip.

Concerning watchOS, I wasn't planning on adding support as part of the integration, I was just pointing out that both of these platforms are unsupported (which turned out to be false, just the deployment targets are not ready yet).

PR #351, opened 2020-07-10
Adding thread safety to MockManager

### LINKED ISSUE #298: Mock Manager Thread Safety
It turns out that now mocked object is not thread safe due to internal mock manager object whose implementation is not thread safe. That can cause problems when one needs to test asynchronous behaviour.

Is there any recommendations how to solve the problem?

One way which I am using currently: generating mocked object, changing implementation to make it thread safe and removing it from build phase regeneration.

In long term I am considering 3 options but all of them requiring fork of the library:

1) Obvious one is to introduce synchronisation mechanisms to MockManager implementation.
2) Create a wrapper inheriting MockManager class and overriding methods with synchronisation feature.
3) Introduce a way to add thread safety to mocked object during generation.

jchitel:
This adds thread safety to a few places inside MockManager, most notably those regions that will get called from actual code, which may be multithreaded.

Regions that are called from tests (creating stubs, verifying, etc.) were left as is because it is not likely that these will be called from multiple threads. Let me know if this is not the case.

Resolves #298

### DIFF
--- a/Source/MockManager.swift
+++ b/Source/MockManager.swift
@@ -25,6 +25,8 @@ public class MockManager {
     private var isSuperclassSpyEnabled = false
     private var isDefaultImplementationEnabled = false
     
+    private let dispatchQueue = DispatchQueue(label: "cuckoo-mock-manager", attributes: .concurrent)
+    
     private let hasParent: Bool
 
     public init(hasParent: Bool) {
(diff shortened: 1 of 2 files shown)


MatyasKriz:
Hey @jchitel, thanks for the contribution!

I'm quite low on time at the moment, so I don't know when I'll be able to do the required research to approve this PR.

In the meantime, could you let me know why you opted for `objc_sync_enter` (and `objc_sync_exit`) instead of utilizing `DispatchQueue`?

Please add some multi-threaded tests that would crash or behave weirdly without the synchronization; one example that is reportedly crashing every time is in the issue this PR has set out to resolve.

jchitel:
Hey, sorry for the delayed response on this.

My understanding of `DispatchQueue` was pretty poor a month ago, but I've since used it a lot more and I see how it can be used here!

I'll make that update and add some tests. Assuming everything behaves as expected, I should have that done today.

jchitel:
Ok, this is done.

I verified that the test fails without the change (though not every time, as is expected from these kinds of issues) and passes with it.

Source/MockManager.swift
```diff
@@ -37,13 +39,15 @@ public class MockManager {
...
+        queue.sync {
+            stubCalls.append(stubCall)
+            unverifiedStubCallsIndexes.append(stubCalls.count - 1)
+        }
 
         if let stub = (stubs.filter { $0.method == method }.compactMap { $0 as? ConcreteStub<IN, OUT> }.filter { $0.parameterMatchers.reduce(true) { $0 && $1.matches(parameters) } }.first) {
             if let action = stub.actions.first {
                 if stub.actions.count > 1 {
-                    stub.actions.removeFirst()
+                    queue.sync { _ = stub.actions.removeFirst() }
```
MatyasKriz, line 50: Hey, consider the scenario of multiple actions on the stub, the testing code looks like this:
```
when(mock.count(characters: any()))
    .thenReturn(0)
    .thenReturn(1)
    .thenReturn(2)
    .thenReturn(3)
```

I'm not very well-versed in testing multi-threaded code, but I like to think that at any given time, there is a thread on every single possible line (and of course some are always hungrily waiting at the mutexes, `sync`s here). That's why at
```
// [0]
if stub.actions.count > 1 {
	// [1]
    queue.sync { _ = stub.actions.removeFirst() }
	// [2]
}
```

[0] **Thread B** is evaluating the condition when `stub.actions.count` is 2, so it goes to the next line and sees that **thread A** has arrived a bit earlier and acquired the mutex.
[1] **Thread A** just acquired the mutex and is about to remove the action, succeeds and brings the state to `stub.actions.count == 1`, the desired state. **Thread A** releases the mutex, having done an honest day's work and continues on with the execution.
[2] However, **thread B** ends its wait for the mutex, acquires it and tries to remove the action; it succeeds, but now the state is `stub.actions.count == 0`, so we're worse off already.
Lo and behold, **thread C** was hiding alongside thread B all along (it passed the condition as well) and when it acquires the mutex and tries to remove an action, the program crashes.

Wrapping the condition in the synchronization fixes this issue, though there might be something else I overlooked:
```
queue.sync {
    if stub.actions.count > 1 {
         stub.actions.removeFirst()
    }
}
```

MatyasKriz:
Also, thanks for bringing my attention to `MockManager.swift`, I noticed a few nitpicks that should have been fixed long ago.

ChaosCoder:
So what's the status of this? We just lost a few hours debugging a test failure in our project and found out, that indeed our problem is exactly that the MockManager isn't thread safe and this PR would fix it. Would love to see this merged! 👍🏻

MatyasKriz:
Hey, @ChaosCoder, the pull request has an unresolved discussion about the implementation, so it's not ready to merge as it is.

MatyasKriz:
I'll close this PR in favor of #392, thanks for your work, @jchitel. 🙂

jchitel:
Sorry for the late response on this. I see the other PR is much more thorough than mine, so it's probably the better bet anyway.

PR #355, opened 2020-07-28
Exits with exit code 1 on input file not found

### LINKED ISSUE #297: Cuckoo doesn't propagate exit status code when generating mocks fails
It seems like running Cuckoo's `generate` command doesn't propagate any status code via `exit` when it fails, and we can't catch any errors during mock generation to, say, stop a build phase or something of the sorts.

meirrosendorff:
Forces the generator to exit with exit code 1 when input files aren't found. This will cause the xcode build to stop if files aren't found.
Resolves #297

### DIFF
--- a/Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
+++ b/Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
@@ -36,7 +36,10 @@ public struct GenerateMocksCommand: CommandProtocol {
         } else {
             inputPathValues = getFullPathSortedArray(options.files)
         }
-        let inputFiles = inputPathValues.map { File(path: $0) }.compactMap { $0 }
+        let inputFiles = inputPathValues.compactMap { File(path: $0) }
+        if inputPathValues.count != inputFiles.count {
+            exit(1)
+        }
         let tokens = inputFiles.map { Tokenizer(sourceFile: $0, debugMode: options.debugMode).tokenize() }
         let tokensWithInheritance = options.noInheritance ? tokens : mergeInheritance(tokens)
 


MatyasKriz:
Hey, @meirrosendorff. Thanks for the PR!

What is the use-case for adding this fatal error when one path that is provided isn't found? I'm not really sure it's a good idea to make something essentially non-optional with this many users. I'd imagine that not everyone keeps their input paths absolutely correct, especially in some of the larger projects I've seen people use Cuckoo in.

However, it's a good idea to check and warn the user about this when we have proper logging in Cuckoo. IMO, an error would be justified if there are paths provided and *none* of them is valid.

Maybe @TadeasKriz can weigh in as well.

meirrosendorff:
Howzit @MatyasKriz 

Thanks for the response, I was having some fun yesterday looking over your code base and I noticed it mentions in the generate_command.feature that when a file isn't found the script should exit with an error code. It seems it doesn't do this because when initializer for `File` fails it prints to the error stream itself without using your `stderrPrint` function so it just carries on. So I thought I would just share a possible solution with you guys. 

In my personal projects, I would definitely prefer that the build crashes on the generator step when a file is not found rather than further on in the build when I try to use the mocks as it narrows down where the problem is and makes the issue more obvious. But I definitely hear your concern that for larger projects a change like this will cause builds to start crashing and be quite a pain.

MatyasKriz:
Sup @meirrosendorff.

Thanks for elaborating on this. I didn't know that this was declared in the manual. I'm but a simple maintainer, so I just learned something new. Though I can't find it, could you please point me to the file where this is located?

It does make sense to me to error out when one of the files is not found, but not by default. The way is to definitely hide it behind a "strict" flag. I like my software to be a little forgiving. Rules (like types) are okay, but when they're too strict, they're IMO at odds with ease of use.

Furthermore, I'd hate to add yet another parameter to the `cuckoo_generator` to modify its behavior. In my opinion, most, if not all, of the parameters belong in a configuration file that wasn't added support for yet.

However, If you'd really like to see this fail-fast possibility included in Cuckoo, adding a `--strict` flag (or some better name if you have one handy) is something I'd merge, because I'm convinced that there are users out there like you that favor minimal number of required paths and would like the generator to tell them which ones are invalid.

PS: Your generated avatar reminds me of the smol Undertale dog.

meirrosendorff:
Hey @MatyasKriz 

I've never looked at my avatar like that and now I can't stop seeing it! 

Its in this file Cuckoo/Generator/Tests/features/generate_command.feature line 50, though I see now that file hasn't been touched for 4 years.

I really like your suggestion when I get some time I'll look at add a flag for it and update this pr.

PR #369, opened 2020-09-15
Add command line option to specific additional frameworks to import into generated files

mwoollard:
This is useful when mocking code from pods imported into the target project. In this case the generated mock won’t be able to access content in the pod. By using the new option you can add import of the pod (or any other) into the generated source file.

### DIFF
--- a/Generator/Source/CuckooGeneratorFramework/FileHeaderHandler.swift
+++ b/Generator/Source/CuckooGeneratorFramework/FileHeaderHandler.swift
@@ -23,12 +23,14 @@ public struct FileHeaderHandler {
         return generationInfo + "\n" + header + "\n"
     }
 
-    public static func getImports(of file: FileRepresentation, testableFrameworks: [String]) -> String {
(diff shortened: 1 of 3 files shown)


MatyasKriz:
Hey, thanks for the PR. I'm not really familiar with the constraints of a `@testable` import as opposed to classic import.
May I know the use-case where adding this argument instead of passing the framework to `--testable` is beneficial? All `--testable` does is it imports the frameworks, just with the `@testable` annotation if I remember correctly.

mwoollard:
@testable elevates access to the imported module so that any internal scope definitions in the module are made public. Personally, I feel its safer, particularly if working within a team, you only expose what should be. Stops someone inadvertently using something from an imported module not directly under test that wouldn't normally be available.

MatyasKriz:
Yes, I'm aware that `@testable import` has this characteristic, though I have to wonder; what role does it play in the generated code as opposed to normal `import`? I can't think of a way the generated code would function differently.

MatyasKriz:
Closing for now, if someone wants to provide an argument for this, I'm all ears.

Issue #159, opened 2017-11-13
Support for nested classes [Feature request]


msewell:
It is common to nest classes for namespacing reasons:

```
// In Outer.swift

class Outer {
    // ...

    class Inner {
        // ...
    }
}
```

Currently, Cuckoo will generate `MockOuter` from Outer.swift, but not `MockInner` (well, `MockOuter.MockInner`). This is a bit of a bummer, since developers will have to either write `MockInnner` themselves, or move `Inner` to the same level as `Outer` in their production code to allow for automated mock generation through Cuckoo, thus negating the benefits of namespacing and polluting the top-level namespace.

Is there a reason Cuckoo cannot generate `MockInner` as well as `MockOuter`? This is a feature I'd really like to see!

michaelom:
Same problem here. 

Having to workaround this limitation is tedious.

karan-sankey:
Also I want to know if there is support for class which inherit from another class(multiple inheritance of class in linear manner).

PR #371, opened 2020-09-19
Nested containers

Tyler-Keith-Thompson:
#159 called for supporting nested classes for namespacing. I too, deeply desired this feature so I spent a few hours getting it in place. The implementation code didn't require a lot of tweaking, it was pretty straightforward. The testing effort was a little dramatic, because there were lots of permutations of things to test. 

To pull this off I introduced a hierarchical concept, "Parent" tokens and "Child" tokens. This is my first contribution to Cuckoo so please forgive any naming/style violations. I tried to make it feel like the rest of the codebase.

### DIFF
--- a/Generator/Source/CuckooGeneratorFramework/Generator.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Generator.swift
@@ -61,7 +61,12 @@ public struct Generator {
         fputs("DECLARATIONS IN MOCK: \(declarations)\n", stdout)
 
         let containers = declarations.compactMap { $0 as? ContainerToken }
-            .filter { $0.accessibility.isAccessible }.map { $0.serializeWithType() }
+            .filter {
+                if let parent = $0.topMostParent {
+                    return parent.accessibility.isAccessible && $0.accessibility.isAccessible
+                }
+                return $0.accessibility.isAccessible
+            }.map { $0.serializeWithType() }
 
         fputs("CONTAINERS IN MOCK: \(containers.map { $0["name"] })\n", stdout)
     
--- a/Generator/Source/CuckooGeneratorFramework/Generator.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Generator.swift
@@ -58,11 +58,26 @@ public struct Generator {
         }
 
         let environment = Environment(extensions: [ext])
-
-        let containers = declarations.compactMap { $0 as? ContainerToken }
+        fputs("DECLARATIONS IN MOCK: \(declarations)\n", stdout)
+        
+        let accessibleDeclarations = declarations.compactMap { $0 as? ContainerToken }
             .filter { $0.accessibility.isAccessible }
-            .map { $0.serializeWithType() }
 
+        let nestedContainers = accessibleDeclarations
+            .filter { $0.accessibility.isAccessible && $0.children.contains { $0 is ContainerToken } }
+            .map { (parent: $0, container:$0.children.compactMap { $0 as? ContainerToken }) }
+            .flatMap { (parent, children) in
+                children.map { child -> [String: Any] in
+                    var c = child
+                    c.parent = parent
+                    return c.serializeWithType()
+                }
+            }
+
+        let containers = nestedContainers + accessibleDeclarations.map { $0.serializeWithType() }
+
+        fputs("CONTAINERS IN MOCK: \(containers.map { $0["name"] })\n", stdout)
+    
         return try environment.renderTemplate(string: Templates.mock, context: ["containers": containers, "debug": debug])
     }
 
--- a/Generator/Source/CuckooGeneratorFramework/Generator.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Generator.swift
@@ -62,8 +62,8 @@ public struct Generator {
 
         let containers = declarations.compactMap { $0 as? ContainerToken }
             .filter {
-                if let parent = $0.topMostParent {
-                    return parent.accessibility.isAccessible && $0.accessibility.isAccessible
+                if let parent = $0.parent {
+                    return parent.allHierarchiesAreAccessible && $0.accessibility.isAccessible
                 }
                 return $0.accessibility.isAccessible
             }.map { $0.serializeWithType() }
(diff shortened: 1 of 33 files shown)


MatyasKriz:
Hey, thanks for the PR!
I have some reservations about the commit messages and some of the naming, but I was unable to do a full review so far. I'll try to get on it as soon as I can!

Tyler-Keith-Thompson:
Appreciate it, I'll be around and fairly responsive if you've got any questions

Generator/Source/CuckooGeneratorFramework/Generator.swift
```diff
@@ -60,9 +60,14 @@ public struct Generator {
...
-            .filter { $0.accessibility.isAccessible }
-            .map { $0.serializeWithType() }
+            .filter {
+                if let parent = $0.parent {
+                    return parent.allHierarchiesAreAccessible && $0.accessibility.isAccessible
+                }
+                return $0.accessibility.isAccessible
+            }.map { $0.serializeWithType() }
 
+
```
MatyasKriz, line 70: I don't think this added line is necessary.

Generator/Source/CuckooGeneratorFramework/Tokens/ParentToken.swift
```diff
@@ -0,0 +1,49 @@
...
+            parent = (p as? ChildToken)?.parent?.value
+        }
+        return true
+    }
+    
+    func adoptAllYoungerGenerations() -> [ParentToken] {
+        return children
+            .compactMap { child -> ParentToken? in
+                guard var c = child as? ContainerToken else { return nil }
+                c.parent = Reference(value: self)
```
MatyasKriz, line 44: What is this initialization of `Reference` trying to achieve? I'd understand creating one instance and then passing it to all the children; however, unless I'm mistaken, this method of use copies (since the declarations are `struct`s) the values for every single child. Was this your intention?

Generator/Source/CuckooGeneratorFramework/Tokens/ContainerToken.swift
```diff
@@ -55,6 +50,8 @@ extension ContainerToken {
             "name": name,
             "accessibility": accessibility.sourceName,
             "isAccessible": accessibility.isAccessible,
+            "hasParent": parent != nil,
+            "parent": parent?.fullyQualifiedName ?? "",
```
MatyasKriz, line 54: I think that `parentFullyQualifiedName` conveys more info than just `parent` and will prevent someone in the future using this property and being confused as to what it really is.

MatyasKriz (changes requested):
The PR seems to be well-put together, though I have some questions and remarks about the code and code style.

When they're addressed, I'll need @TadeasKriz to also review all of this before merging.

Generator/Source/CuckooGeneratorFramework/Tokens/StructDeclaration.swift
```diff
@@ -0,0 +1,23 @@
...
+    // NOTE: Purely for supporting nested classes, could be any generic parent (like extensions)
+    public let name: String
+    public var accessibility: Accessibility
+    public let range: CountableRange<Int>
+    public let nameRange: CountableRange<Int>
+    public let bodyRange: CountableRange<Int>
+    public let children: [Token]
+    
+    public func isEqual(to other: Token) -> Bool {
+        guard let other = other as? ExtensionDeclaration else { return false }
```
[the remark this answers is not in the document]
Tyler-Keith-Thompson: This is actually a protocol requirement for a `Token` I had taken the implementation from `ExtensionDeclaration`. Instead of range I could change it to use `name` like `ClassDeclaration` does, though.

Generator/Source/CuckooGeneratorFramework/Tokens/ParentToken.swift
Tyler-Keith-Thompson: Good catch. Went through all that trouble turning into a reference type that I then didn't use correctly :/

Generator/Source/CuckooGeneratorFramework/Tokens/ParentToken.swift
```diff
@@ -38,10 +38,11 @@ extension ParentToken {
     }
     
     func adoptAllYoungerGenerations() -> [ParentToken] {
+        let parentReference:Reference<ParentToken> = Reference(value: self)
```
MatyasKriz, line 41: Is the type declaration necessary? And if so, please pad it with a single space as per my previous comment.

Tyler-Keith-Thompson: Unfortunately it is necessary, otherwise it implicitly resolves to `Reference<Self>` which *should* be fine but the compiler doesn't like it, so here we are.

I've added the space as requested. Have you guys considered adding a linter to the project?

Generator/Source/CuckooGeneratorFramework/Tokens/StructDeclaration.swift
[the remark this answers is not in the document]
Tyler-Keith-Thompson: What would you like done here?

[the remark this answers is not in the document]
MatyasKriz: Well, I'm not sure where this `isEqual(to:)` is used, but I reckon the method needs to be reliable, so yes, I'd change it to compare `name`, but most importantly not compare type to `ExtensionDeclaration` but `StructDeclaration`.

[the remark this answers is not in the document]
Tyler-Keith-Thompson: Alright I've fixed the obvious things about this. For this PR I'm hoping that's good enough. Turns out the place this ends up being used is in `FileRepresentation.swift` and is part of the `mergeInheritence` function.

It seems like there could be a reasonable refactor of that to be more clear and to only apply equality to token types that can be inherited. If it would be helpful/desired I'm more than happy to create a separate PR to address that.

MatyasKriz:
I checked this out yesterday on my machine and the tests seem to cover the use-cases well, so now @TadeasKriz needs to review this and we can merge it.

Tyler-Keith-Thompson:
Hey @TadeasKriz now that it's the new year, have you had a chance to look at this?

SaezChristopher:
up

MatyasKriz:
I decided to stop waiting and just merge this. 😄 We'll just see how it goes, thanks again for the PR, I'll release it shortly.

PR #374, opened 2020-12-08
Support calling objective-c methods and closures with nil

dekpient:
Hello, I'm trying to fix 2 issues I found when trying to use `Cuckoo/OCMock` in a project. I'm currently unable to cover a few code branches because of them.

1. Calling `mock.say(nil)` currently fails with `-[_NSArrayM insertObject:atIndex:]: object cannot be nil (NSInvalidArgumentException)` because `NSInvocation+OCMockWrapper` doesn't do a null check.
2. Using `objectiveArgumentClosure` and calling the closure with `nil` fails with `-[NSNull count]: unrecognized selector sent to instance 0x7fff8002ebb0 (NSInvalidArgumentException)`. I think it's cause `TrustMe` doesn't support nullable types. I propose `objectiveOptionalArgumentClosure` (or maybe `objcArgumentClosureWithOptional` ?). I can add more of them to support closures with more args if you think this is alright.

I also removed the reference to `ObjectiveExamplesTest.swift`. It doesn't exist in the project so it's making the build fail.

Cheers

### DIFF
--- a/OCMock/ObjectiveArgumentClosure.swift
+++ b/OCMock/ObjectiveArgumentClosure.swift
@@ -31,6 +31,19 @@ public func objectiveArgumentClosure<IN1, IN2>(from: Any) -> (IN1, IN2) -> Void 
     }
 }
 
+// TODO add more variants
+public func objectiveOptionalArgumentClosure<IN1, IN2>(from: Any) -> (IN1?, IN2?) -> Void {
+    return { in1, in2 in
+        var arg = from
+        let block = UnsafeRawPointer(&arg).assumingMemoryBound(to: (@convention(block) (NSObject?, NSObject?) -> Void).self).pointee
+
+        let nsIn1 = in1 == nil ? nil : TrustMe<NSObject>.onThis(in1 as Any)
+        let nsIn2 = in2 == nil ? nil : TrustMe<NSObject>.onThis(in2 as Any)
+
+        block(nsIn1, nsIn2)
+    }
+}
+
 public func objectiveArgumentClosure<IN1, IN2, IN3>(from: Any) -> (IN1, IN2, IN3) -> Void {
     return { in1, in2, in3 in
         var arg = from
--- a/Tests/OCMock/ObjectiveClassTest.swift
+++ b/Tests/OCMock/ObjectiveClassTest.swift
@@ -160,6 +160,39 @@ class ObjectiveClassTest: XCTestCase {
 
         objcVerify(mock.dudka(lelo: objcAny()))
     }
+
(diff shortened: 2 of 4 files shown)


MatyasKriz:
Hey, @dekpient. Thanks for the PR and thanks for using the OCMock integration!

It's a nice find with the `nil`, I'll take a look at the code and see what we can do with the second problem you're proposing a solution to.

OCMock/ObjectiveArgumentClosure.swift
```diff
@@ -31,6 +31,19 @@ public func objectiveArgumentClosure<IN1, IN2>(from: Any) -> (IN1, IN2) -> Void
     }
 }
 
+// TODO add more variants
+public func objectiveOptionalArgumentClosure<IN1, IN2>(from: Any) -> (IN1?, IN2?) -> Void {
```
MatyasKriz, line 35: I think we'll have to try coming up with a more complex solution to this, because as the number of arguments go up, the combinations of optional and non-optional arguments will skyrocket.

Tests/OCMock/ObjectiveClassTest.swift
```diff
@@ -160,6 +160,39 @@ class ObjectiveClassTest: XCTestCase {
...
+        }
+
+        mock.say("cheeze") { result, messages in
+            XCTAssertEqual("meh", result)
+            XCTAssertNil(messages)
+            expectCompleteWithNil.fulfill()
+        }
+
+        mock.say(nil) { result, messages in
+            XCTAssertEqual("nil now behaves", result)
```
MatyasKriz, line 186: Good `nil`.

OCMock/ObjectiveArgumentClosure.swift
dekpient: Yeah, I agree. I ended up defining this in my test project just for the specific test cases that I need. Should I split this into 2 PRs so the 1st problem can be fixed? I really need that fix.

MatyasKriz: Splitting this up sounds like the better call at the moment. Would you please add the remaining combinations for 2-parameter closures, i.e. (optional, non-optional), (non-optional, optional) as well as an optional counterpart to the 1-parameter closure. We'll need to come up with a more generic solution, but for now this can help make the closures more accessible by showing users how to implement them (only using the right amount of arguments). Thanks!

Also, would you mind creating a test or two for the closures with optional parameters? Just to make sure they work and when replacing with the generic solution to know nothing has been broken.

Afterwards, we can probably merge this and release.

PR #392, opened 2021-05-11
Make stubbed calls more thread safe

### LINKED ISSUE #298: Mock Manager Thread Safety

ChaosCoder:
This PR adds thread safety around memorizing stub calls inside the MockManager.

Calling stubbed methods from a multithreading environment lead to crashing. By using a private queue for synchronizing write access to the appropriate collections calling stubs multithreaded is no longer a problem.

Resolves #298 and supersedes #351.

### DIFF
--- a/Source/MockManager.swift
+++ b/Source/MockManager.swift
@@ -45,14 +45,10 @@ public class MockManager {
         }
 
         if let stub = (stubs.filter { $0.method == method }.compactMap { $0 as? ConcreteStub<IN, OUT> }.filter { $0.parameterMatchers.reduce(true) { $0 && $1.matches(parameters) } }.first) {
-            var action: StubAction<IN, OUT>?
-            queue.sync {
-                action = stub.actions.first
-                if stub.actions.count > 1 {
-                    stub.actions.removeFirst()
-                }
-            }
-            guard let action = action else {
+            
+            guard let action = queue.sync(execute: {
+                return stub.actions.count > 1 ? stub.actions.removeFirst() : stub.actions.first
+            }) else {
                 failAndCrash("Stubbing of method `\(method)` using parameters \(parameters) wasn't finished (missing thenReturn()).")
             }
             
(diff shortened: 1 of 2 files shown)


Source/MockManager.swift
```diff
@@ -37,33 +39,39 @@ public class MockManager {
...
-        unverifiedStubCallsIndexes.append(stubCalls.count - 1)
+        queue.sync {
+            stubCalls.append(stubCall)
+            unverifiedStubCallsIndexes.append(stubCalls.count - 1)
+        }
 
         if let stub = (stubs.filter { $0.method == method }.compactMap { $0 as? ConcreteStub<IN, OUT> }.filter { $0.parameterMatchers.reduce(true) { $0 && $1.matches(parameters) } }.first) {
-            if let action = stub.actions.first {
+            var action: StubAction<IN, OUT>?
+            queue.sync {
```
MatyasKriz, line 49: With this configuration I think it's better to change the variable into a constant like so:
```
let action = queue.sync {
  if !stub.actions.isEmpty {
    return stub.actions.removeFirst()
  } else {
    return nil
  }
}
```

Source/MockManager.swift
```diff
@@ -75,29 +83,35 @@ public class MockManager {
...
-                    return try implementation(parameters)
-                case .returnValue(let value):
-                    return value
-                case .throwError(let error):
-                    throw error
-                case .callRealImplementation where hasParent:
-                    return try superclassCall()
-                default:
-                    failAndCrash("No real implementation found for method `\(method)`. This is probably caused  by stubbed object being a mock of a protocol.")
+                    queue.sync { _ = stub.actions.removeFirst() }
```
MatyasKriz, line 96: Why does it need to be inside another `queue.sync` here?

MatyasKriz (changes requested):

MatyasKriz:
Thanks for the PR, @ChaosCoder! It looks good, although it can be a bit better. 🙂

Source/MockManager.swift
ChaosCoder: Your suggestion actually changes the logic (in the case of `stub.actions.count == 1`, but I agree, the constant is nicer. I pushed another commit.

Source/MockManager.swift
ChaosCoder: Good catch, that was an error.

Source/MockManager.swift
MatyasKriz: Yeah, my bad, sorry. Didn't read the code properly.

MatyasKriz:
Thanks for resolving the issues so quickly! Looks great and I will try to merge this as soon as I can. 🙂

MatyasKriz (approved):

MatyasKriz:
Thanks again for the PR! I've merged it and will release it shortly.

ChaosCoder:
So, just to make sure this does not get lost: this PR is still not merged. ;)

Issue #407, opened 2021-11-30
Async/await support


austinrovge:
Now that Swift 5.5 is released and async/await will be back deployed to iOS 13 in Xcode 13.2, I'm getting ready to update my app to take advantage of it.

Is there any ongoing or planned work for adding support for async functions/properties/initializers, actors, etc? I'm interested in contributing to this feature, though I don't have a lot of free time at the moment.

MatyasKriz:
Hey, @austinrovge, I was thinking about this blocking users as well. Unfortunately I'm not strong in the free time department either.

SourceKitten seems to support this since [0.31.1](https://github.com/jpsim/SourceKitten/releases/tag/0.31.1), so we should be able to get this information pretty easily once it's updated. If you're up for it, by all means, go for implementing this. 🙂 Please include a few simple test cases along with it.

PR #410, opened 2022-01-17
Add support for concurrency mocking

ailtonvivaz:
This PR implements support for parsing async methods and closures  as discussed in #407

### DIFF
--- a/Generator/Source/CuckooGeneratorFramework/Templates/NopImplStubTemplate.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Templates/NopImplStubTemplate.swift
@@ -8,7 +8,10 @@
 extension Templates {
     static let noImplStub = """
 {{container.accessibility}} class {{ container.name }}Stub{{ container.genericParameters }}: {{ container.name }}{% if container.isImplementation %}{{ container.genericArguments }}{% endif %} {
-    {% for property in container.properties %}
+    {% for property in container.properties %}    
+    {% if debug %}
+    // {{property}}
+    {% endif %}
     {% for attribute in property.attributes %}
     {{ attribute.text }}
     {% endfor %}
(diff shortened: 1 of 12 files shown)


Generator/Source/CuckooGeneratorFramework/Templates/NopImplStubTemplate.swift
```diff
@@ -31,6 +34,12 @@ extension Templates {
     {% endfor %}
 
     {% for method in container.methods %}
+    {% if debug %}
+    // {{method}}
+    {% endif %}
+    {% for attribute in method.attributes %}
```
MatyasKriz, line 40: What do these attributes do and how are they linked to async/await?

Tests/Swift/Source/TestedProtocol.swift
```diff
@@ -17,7 +17,7 @@ protocol TestedProtocol {
     func noReturn()
 
     func count(characters: String) -> Int
-
+
```
MatyasKriz, line 20: Please remove this whitespace. 🙂

MatyasKriz (commented):
What a superb pull request! Well done, @ailtonvivaz. I have a minor question, but after that we can merge this immediately. I think it warrants its own release as this is quite a good feature to have, we just weren't able to find the time, so thank you very much!

Generator/Source/CuckooGeneratorFramework/Templates/NopImplStubTemplate.swift
ailtonvivaz: The concurrency in Swift is available just for iOS 13+, macOS 10.15+,watchOS 6.0+ and tvOS 13.0+.
As the library works with iOS 8+, macOS 10.9+ and tvOS 9+, some users could use the @available attribute to provide some methods with this capability.
And, as `NopImplStubTemplate.swift` hadn't this implementation yet, I've added.

Tests/Swift/Source/TestedProtocol.swift
ailtonvivaz: Thanks. It was removed

Generator/Source/CuckooGeneratorFramework/Templates/NopImplStubTemplate.swift
MatyasKriz: I see, thanks for explaining. 🙂

MatyasKriz:
I've merged this into `master`, but I can't seem to be able to build it with the current values in the `@available` attribute, Xcode keeps suggesting to raise them. I'll update Xcode to the latest version and try again.

ailtonvivaz:
> I've merged this into `master`, but I can't seem to be able to build it with the current values in the `@available` attribute, Xcode keeps suggesting to raise them. I'll update Xcode to the latest version and try again.

Maybe your Xcode version is below 13.2. Before this, concurrency was supported just for iOS15+.
https://developer.apple.com/documentation/xcode-release-notes/xcode-13_2-release-notes

Updating to the new Xcode, you'll get this feature.

Issue #328, opened 2020-01-24
Escape reserved keywords


paiv:
Trying to mock protocols with escaped reserved keywords:
```swift
protocol Example {
    func `nil`() -> Void
}
```

Generates a mock with compilation errors like:

```swift
     func nil()  {
//        ^
// Keyword 'nil' cannot be used as an identifier here
```

mwoollard:
I'm effected by this so am going to take a look at resolving.

MatyasKriz:
Thanks to @sk409 for implementing this! This change will be included in the next release, though I'll close the issue so I don't forget. 🙂

PR #412, opened 2022-02-05
Escape reserved keywords.

sk409:
I escaped reserved keywords which is not allowed to be used as method name to fix this issue(#328)

This is my first PR, if I made a mistake, please point it out.


MatyasKriz:
Merged. 🙂 Thanks for the pull request! It will be included in the next release.

PR #415, opened 2022-02-14
Ignore missing input files.

### LINKED ISSUE #404: Missing input paths lead to messed up output files
When a user provides input paths to missing files and output into multiple files is active,
the generator will mess up the output files.

## Example

Existing files: A, C
User Input: A, B, C
Generated Output files: MockA, MockB

MockB will contain output for C.
MockC will be missing.

Bug is located in `GenerateMocksCommand` as can be seen in the following excerpt.
The output loop uses `inputPathValues` as source for output paths.
But `inputPathValues` is unfiltered. Actually used files are located in `inputFiles`.

Snippet from line 72 - 77:

```swift
let inputPaths = inputPathValues.map { Path($0) }
for (inputPath, outputText) in zip(inputPaths, mergedFiles) {
    let fileName = options.filePrefix + inputPath.fileName
    let outputFile = TextFile(path: outputPath + fileName)
    try outputText |> outputFile
}
```

## Solution proposals

a. Throw an error for invalid input/missing files (breaking change)
b. Fix output paths, e.g. by using inputFiles as a source for paths.

sk409:
Fixes #404 

The cause of this issue is that inputPathValues that is not filtered is used to determine file name while mergedFile is filtered.

I used inputFiles as a source of inputPaths instead of inputPathValues.

Because inputFiles is filtered to ignore missing files and mergedFiles is made from inputFiles.

### DIFF
--- a/Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
+++ b/Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
@@ -63,7 +63,7 @@ public struct GenerateMocksCommand: CommandProtocol {
 
         do {
             if outputPath.isDirectory {
-                let inputPaths = inputFiles.map { Path($0.path!) }
+                let inputPaths = inputFiles.compactMap { $0.path }.map { Path($0) }
                 for (inputPath, outputText) in zip(inputPaths, mergedFiles) {
                     let fileName = options.filePrefix + inputPath.fileName
                     let outputFile = TextFile(path: outputPath + fileName)
--- a/Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
+++ b/Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
@@ -63,7 +63,7 @@ public struct GenerateMocksCommand: CommandProtocol {
 
         do {
             if outputPath.isDirectory {
-                let inputPaths = inputPathValues.map { Path($0) }
+                let inputPaths = inputFiles.map { Path($0.path!) }
                 for (inputPath, outputText) in zip(inputPaths, mergedFiles) {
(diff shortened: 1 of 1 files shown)


Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
```diff
@@ -63,7 +63,7 @@ public struct GenerateMocksCommand: CommandProtocol {
 
         do {
             if outputPath.isDirectory {
-                let inputPaths = inputPathValues.map { Path($0) }
+                let inputPaths = inputFiles.map { Path($0.path!) }
```
MatyasKriz, line 66: What do you think about `inputFiles.compactMap { $0.path }.map(Path.init)` instead? I don't like the look of the force unwrap.

sk409: Thank you for your review.
I thought it was safe to use force unwrapping because the inputFilesFile was initialized using File (path: String), but I agree with your opinion that you don't like force unwrapping.

sk409: I tried to use `Path.init` but Xcode show me an error said "Path is ambiguous", so I used `Path($0` instead.

MatyasKriz: I see, that's okay, the current solution is in no way inferior. 🙂

MatyasKriz:
Thanks for yet another great pull request! I know this one was a bit on the simpler side, but it's important nonetheless. I might not have the time to fix the issues, but I can at least merge the efforts of contributors like you. 🙂

I'll merge this shortly, though it will not be released yet.

MatyasKriz (approved):

PR #416, opened 2022-02-15
OCMock with Swift Package Manager

### LINKED ISSUE #335: OCMock with Swift Package Manager
Is there a plan to use OCMock with Swift Package Manager? Would be awesome to stub Apples Objective-C classes with Cuckoo via SPM 😀

ajpallares:
Fixes #335.

Note that because OCMock contains `unsafeOptions`, the dependency of OCMock has to be specified by a commit hash (see https://github.com/erikdoe/ocmock/pull/460#issuecomment-720786603).

### DIFF
--- a/OCMock/Swift/NSValueConvertible.swift
+++ b/OCMock/Swift/NSValueConvertible.swift
@@ -7,9 +7,6 @@
 
 import Foundation
 import CoreGraphics
-#if os(iOS)
-import UIKit
-#endif
 
 public protocol NSValueConvertible {
     func toNSValue() -> NSValue
--- a/OCMock/Swift/NSValueConvertible.swift
+++ b/OCMock/Swift/NSValueConvertible.swift
@@ -7,6 +7,9 @@
 
 import Foundation
 import CoreGraphics
+#if os(iOS)
+import UIKit
+#endif
 
 public protocol NSValueConvertible {
     func toNSValue() -> NSValue
--- a/OCMock/Swift/ObjectiveMatchers.swift
+++ b/OCMock/Swift/ObjectiveMatchers.swift
@@ -6,9 +6,6 @@
 //
 
 import Foundation
-#if SWIFT_PACKAGE
-import Cuckoo_OCMock_Objc
-#endif
 
 /// Used as an Objective-C matcher matching any Objective-C closure.
 public func objcAny<T: NSObject>() -> T {
--- a/OCMock/Swift/ObjectiveMatchers.swift
+++ b/OCMock/Swift/ObjectiveMatchers.swift
@@ -6,6 +6,9 @@
 //
 
 import Foundation
+#if SWIFT_PACKAGE
+import Cuckoo_OCMock_Objc
+#endif
 
 /// Used as an Objective-C matcher matching any Objective-C closure.
 public func objcAny<T: NSObject>() -> T {
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:4.0
+// swift-tools-version:5.0
 
 import PackageDescription
 
--- a/Package.swift
+++ b/Package.swift
@@ -8,12 +8,30 @@ let package = Package(
         .library(
             name: "Cuckoo",
             targets: ["Cuckoo"]),
+        .library(
+            name: "Cuckoo_OCMock",
+            targets: ["Cuckoo", "Cuckoo_OCMock"]),
+    ],
+    dependencies: [
+        .package(
+            url: "https://github.com/erikdoe/ocmock",
+            .revisionItem("afd2c6924e8a36cb872bc475248b978f743c6050")
+        )
     ],
     targets: [
         .target(
             name: "Cuckoo",
             dependencies: [],
             path: "Source"),
+        .target(
+            name: "Cuckoo_OCMock-Objc",
+            dependencies: ["OCMock"],
+            path: "OCMock/ObjCHelpers",
+            publicHeadersPath: "."),
+        .target(
+            name: "Cuckoo_OCMock",
+            dependencies: ["Cuckoo", "Cuckoo_OCMock-Objc"],
+            path: "OCMock/Swift"),
         .testTarget(
             name: "CuckooTests",
             dependencies: ["Cuckoo"],
--- a/Cuckoo.xcodeproj/project.pbxproj
+++ b/Cuckoo.xcodeproj/project.pbxproj
@@ -7,7 +7,6 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		000373309CA86A4795E141D0 /* NSValueConvertible.swift in Sources */ = {isa = PBXBuildFile; fileRef = 989EFADEC10AB6D627A5FAE6 /* NSValueConvertible.swift */; };
 		0005B8933E95841966E75631 /* MockManager+preconfigured.swift in Sources */ = {isa = PBXBuildFile; fileRef = E83C4A633699E5B88D9A9C1B /* MockManager+preconfigured.swift */; };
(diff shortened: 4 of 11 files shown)


OCMock/Swift/ObjectiveMatchers.swift
```diff
@@ -6,6 +6,9 @@
 //
 
 import Foundation
+#if SWIFT_PACKAGE
```
MatyasKriz, line 9: I don't really like how the code is changed just to work with a different dependency manager. I found something that Firebase uses https://github.com/firebase/firebase-ios-sdk/blob/master/HeadersImports.md#header-file-types-and-locations---for-header-file-creators, though I'm not sure if we can go that way as well.

What do you think? If changing the project structure is what it takes, it's a small price to pay to keep the files consistent.

MatyasKriz (commented):
I'm mostly new to SPM and curious as to how it works, so please excuse my questions if they seem too stupid. As for the `#if` directives, those seem a bit painful for me, especially as I expect to not get a feedback that I forgot one until after I try to do some SPM work, correct?

Package.swift
```diff
@@ -8,12 +8,30 @@ let package = Package(
...
+            .revisionItem("afd2c6924e8a36cb872bc475248b978f743c6050")
+        )
     ],
     targets: [
         .target(
             name: "Cuckoo",
             dependencies: [],
             path: "Source"),
+        .target(
+            name: "Cuckoo_OCMock-Objc",
```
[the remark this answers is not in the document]
ajpallares: > Is this target separate just because it's on different path or is there a different reason?

Yes. The reason is that it's not desirable for people who only need Cuckoo to also have to include the Cuckoo/OCMock subdependency (which also includes OCMock itself). This way, when we someone wants to add Cuckoo to their project, Xcode be prompts for which of the two versions to use. For example, in [OHHTTPStubs](https://github.com/AliSoftware/OHHTTPStubs/blob/master/Package%40swift-5.swift) this prompt appears:
image
And this is the products this package exposes
```
    products: [
        .library(
            name: "OHHTTPStubs",
            targets: [
                "OHHTTPStubs",
            ]
        ),
        .library(
            name: "OHHTTPStubsSwift",
            targets: [
                "OHHTTPStubs",
                "OHHTTPStubsSwift"
            ]
        )
    ],
```
 
> I also see that this target passes the publicHeadersPath argument, what is the magic in that?

By default, if not specified, "the path to the directory containing public headers of a C-family target" ([official documentation](https://developer.apple.com/documentation/swift_packages/target/2880338-publicheaderspath)), the path is assumed to be `include`. I thought that, rather than relocating the headers into an `include` folder, it would be better to simply make public all the headers in the root folder, which is achieved by setting `publicHeadersPath` to `"."`

OCMock/Swift/NSValueConvertible.swift
```diff
@@ -7,6 +7,9 @@
 
 import Foundation
 import CoreGraphics
+#if os(iOS)
+import UIKit
```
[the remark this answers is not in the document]
ajpallares: Because this file uses UIKit types
image
That said, maybe there's a way to avoid the import. I don't know how this is achieved with CocoaPods 🤔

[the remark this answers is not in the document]
ajpallares: In fact, I'm thinking that rather than `#if os(iOS)`, perhaps the condition should be `#if canImport(UIKit)`. But anyways, later in the file the condition used is `#if os(iOS)` so, it wouldn't make any difference in this case

OCMock/Swift/ObjectiveMatchers.swift
ajpallares: I've found an alternative [navigating through Firebase's code](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCombineSwift/Sources/Core/Core.swift). And it's the use of the undocumented `@_exported import`. It's even mentioned by a Developer Tools Engineer at  [here](https://developer.apple.com/forums/thread/650995). Here's [an explanation in the official Swift forum](https://forums.swift.org/t/what-does-exported-import-do/35869).

In our case, we would have a new Swift file with only this code:
```
#if SWIFT_PACKAGE
@_exported import Cuckoo_OCMock_Objc
@_exported import Cuckoo
#if os(iOS)
@_exported import UIKit
#endif
#endif
```

This way we achieve two things:

* Remove the need to `import UIKit` in `NSValueConvertible.swift` (comment above) and all the `import Cuckoo_OCMock_Objc`.
* When integrating Cuckoo+OCMock, it would be enough to `import Cuckoo_OCMock` (with the solution in the current PR, it would be required to import both `Cuckoo` and `Cuckoo_OCMock`).

However, this has one disadvantage. We would be exposing the Objective-C bridge APIs, which is something that is not desired (correct me if I'm wrong).

What do you think about this `@_exported` option? Anyway, I'm going to keep investigating the Firebase link you shared.

ajpallares:
Don't worry! I'm also rather new to SPM but the good thing about this is that we learn from it 😊

In any case, I didn't find any of your questions to be stupid at all! I think it's a good thing that you care a lot about this library and that you require minimum quality standards.

> As for the #if directives, those seem a bit painful for me, especially as I expect to not get a feedback that I forgot one until after I try to do some SPM work, correct?

Totally agree. They are painful and easy to be forgotten in the future.

> However, I do appreciate the PR and want to add support for SPM, I wouldn't want to block anyone from switching to their dependency manager of choice.

No problem. We'll get it working for sure. I'm already using Cuckoo with SPM in our project but I needed to add some ObjC mocks and the straightforward solution is to use Cuckoo + OCMock. So **thank you** for this amazing library! 🙌

OCMock/Swift/NSValueConvertible.swift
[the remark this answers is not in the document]
MatyasKriz: I see, that's curious.

OCMock/Swift/ObjectiveMatchers.swift
MatyasKriz: From what I gather, the `@_exported import` seems easier to use that what the Firebase proposes, so at least for now I think it's the better option.

The disadvantage is, as I understand it, that the user's test code will have access to types like `TrustMe`, `TrustHim` and so on, which are used internally? I don't feel like it's a dealbreaker, especially as it's better than no OCMock support in SPM or the alternative of the `#if` directive in every relevant file.

Is it possible to diminish the disadvantage by wrapping the code you proposed in another compiler directive that would only call these imports when OCMock is available? Or I suppose if the file is in the correct targets, SPM will deal with this automatically, right?

ajpallares: > The disadvantage is, as I understand it, that the user's test code will have access to types like TrustMe, TrustHim and so on, which are used internally?

That's correct.

> Is it possible to diminish the disadvantage by wrapping the code you proposed in another compiler directive that would only call these imports when OCMock is available? Or I suppose if the file is in the correct targets, SPM will deal with this automatically, right?

Since the file with these imports will only belong to the `Cuckoo_OCMock` target, it will only be available for those devs that use `Cuckoo_OCMock`. If you only need Cuckoo without OCMock, then things remain the same. You simply `import Cuckoo` and you won't have access to either OCMock or anything from the Cuckoo+OCMock bridge.

MatyasKriz: That's good enough for me. Great work, @ajpallares, and thanks for the PR! I'll just do some testing and merge this as soon as I can. 🙂

MatyasKriz:
Merged! Thanks a lot, @ajpallares, for this great PR and the research that you've put into making Cuckoo+OCMock work with SPM. 🙂

PR #418, opened 2022-03-07
feat: enable OCMock's `OCMQuantifier` for `objcVerify` function. Fixes #366

### LINKED ISSUE #366: OCMock - How to verify that a method has been called a certain number of times
Currently I wanna make sure that the method `numberOfSections(in:)` just called only once time, any posibilty to make it as `objcVerify(mock, times(1)).numberOfSections(in: tableView))`? thanks 🙏

```swift
// stubbing the class is very similar to stubbing with Cuckoo
let mock = objcStub(for: UITableViewController.self) { stubber, mock in
  stubber.when(mock.numberOfSections(in: tableView)).thenReturn(1)
}

objcVerify(mock.numberOfSections(in: tableView))
```

ajpallares:
Added support for `OCMQuantifier` to `objcVerify` function.

This PR fixes #366

### DIFF
--- a/OCMock/ObjectiveVerify.swift
+++ b/OCMock/ObjectiveVerify.swift
@@ -6,14 +6,16 @@
 //
 
 import XCTest
+import OCMock
 
 extension XCTestCase {
     /**
      * Objective-C equivalent to Cuckoo's own `verify`.
      * - parameter invocation: Autoclosured invocation of the method/variable that is being verified.
+     * - parameter quantifier: Verification to assert how many times the call was made.
      */
-    public func objcVerify<OUT>(_ invocation: @autoclosure () -> OUT, file: StaticString = #file, line: UInt = #line) {
-        OCMMacroState.beginVerifyMacro(at: OCMLocation(testCase: self, file: String(file), line: line))
+    public func objcVerify<OUT>(_ invocation: @autoclosure () -> OUT, _ quantifier: OCMQuantifier = .exactly(1), file: StaticString = #file, line: UInt = #line) {
+        OCMMacroState.beginVerifyMacro(at: OCMLocation(testCase: self, file: String(file), line: line), with: quantifier)
         _ = invocation()
         OCMMacroState.endVerifyMacro()
     }


ajpallares:
Even though OCMock verifies by default that the method has been called at least once (see Quantifiers in [OCMock docs](https://ocmock.org/reference/#verifying-interactions)), I set the default `OCMQuantifier` to `.exactly(1)` in `objcVerify` in order for it to behave the same as Cuckoo's `verify`, which defaults to the `times(1)` `CallMatcher`.

Let me know if this needs to be changed.

ajpallares:
I'm not sure about the default `OCMQuantifier` being `.exactly(1)`. There's no doubt it is consistent with Cuckoo's default `CallMatcher` also being `times(1)`. However, prior to this PR, `objcVerify` was actually checking against `.atLeast(1)` quantifier. 

Because of this, existing tests might need to change after this PR, which, indeed, was necessary for the tests in Cuckoo (see [this commit](https://github.com/Brightify/Cuckoo/pull/418/commits/847abe5cf21154e78d5031d37634702df8141166)). It can be annoying and confusing for Cuckoo users to realize tests are failing after updating Cuckoo's version. 

WDYT @MatyasKriz?

rolandkakonyi:
@MatyasKriz @ajpallares did you think about merging this PR in a minor version with the slight change of matching the current behavior?

Avoiding the behavior change could be done in 2 different ways in my opinion:
- A: setting the default value to `.atLeast(1)`
- B: define `objcVerify` two times, once without a `quantifier` for keeping the current behavior intact and once with `quantifier` where one can explicitly define it

The outcome of both would be the very same and would allow shipping it in a minor version update.

Another PR could be opened with the current implementation of the PR where the default value matches to `CallMatcher` and would be shipped in next major version update.

MatyasKriz:
>     * A: setting the default value to `.atLeast(1)`

I suppose that this is a nice enough solution to deliver the core feature of this PR. I'd then immediately create another PR that would change this to `.exactly(1)` that we could then merge for the next major update. 🙂

MatyasKriz:
@ajpallares I've rebased the relevant changes according to @rolandkakonyi's suggestion while leaving the PR that makes the quantifier consistent with Cuckoo's verify for when the next major version is released.
This PR is already merged, so it can be closed and I'll release a new version in a minute. 🙂

ajpallares:
That’s great! Thank you all 🙌🏼

Issue #419, opened 2022-03-09
Unterminated string literal when mocking function with long callback


bbrk24:
I'm mocking a protocol with a function that looks like this:

```swift
func foo(
    completion: @escaping (
        _ bar: Int,
        _ baz: Int,
        _ qux: Int
    ) -> Void
)
```

The generated mocks do not properly remove the line breaks before attempting to put it into a single-line string literal, resulting in the errors `Unterminated string literal`, `Consecutive statements on a line must be separated by ';'`, `Expected expression`, and `'_' can only appear in a pattern or on the left side of an assignment`, several times each.

PR #424, opened 2022-04-30
Print method name using multiline string

sk409:
### Summary

I printed method name using multiline string to fix this issue(#419).

### Problem

Closure parameter separated into multi line had been printed without any transformation.

This led to compilation error like this:

```swift
"foo(completion: @escaping (
            _ bar: Int,
            _ baz: Int,
            _ qux: Int
        ) -> Void)"
```

My solution is to use multiline string liike this:
```swift
"""
    foo(completion: @escaping (
            _ bar: Int,
            _ baz: Int,
            _ qux: Int
        ) -> Void)
    """
```

Thank you for your confirmation.

### DIFF
--- a/Generator/Source/CuckooGeneratorFramework/Templates/MockTemplate.swift
+++ b/Generator/Source/CuckooGeneratorFramework/Templates/MockTemplate.swift
@@ -104,7 +104,9 @@ extension Templates {
     {% endfor %}
     {{ method.accessibility }}{% if container.isImplementation and method.isOverriding %} override{% endif %} func {{ method.name }}{{ method.genericParameters }}({{ method.parameterSignature }}) {{ method.returnSignature }} {
         {{ method.self|openNestedClosure }}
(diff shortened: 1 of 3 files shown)


MatyasKriz (approved):
The changes look good to me, however, would you please add some multiline method to some test class, so that the compiler can verify this as well? 🙂

sk409:
I added some method to test multiline argument method.
If I should fix something, please let me know.

sk409:
I now realize that my solution only works for users who are using space for indentation.😢

MatyasKriz:
> I now realize that my solution only works for users who are using space for indentation.😢

Is there an easy way to fix that? Otherwise I wouldn't dwell on it too much. 😄

MatyasKriz:
Thanks a lot for the PR! I don't think the space problem is too bad, so I'll merge this now. 🙂

sk409:
Thank you for merging.
I think my solution works for those who use space as indentation.

PR #430, opened 2022-08-05
Remove unnecessary await expression in MockTemplate

uebo:
Hi, the following warning occurred due to Swift Concurrency support.

`No 'async' operations occur within 'await' expression`

Screen Shot 2022-08-05 at 14 18 17

I think that `crashOnProtocolSuperclassCall()` method is not an async method so I don't need an await clause.

Thanks.


MatyasKriz:
I've added some async methods and it does indeed seem like the await is superfluous in this case. Thanks a lot for your contribution. 🙂 I've merged it and it will be released in the next version shortly.

PR #431, opened 2022-08-16
Breaking change: Make OCMock's default OCMQuantifier consistent to Cuckoo's

MatyasKriz:
This PR breaks the API and as such can't be released in a non-major version.


MatyasKriz:
Merged into SwiftSyntax branch.

Issue #394, opened 2021-06-28
Closure variable generates with error


evgzor:
Xcode 12.5.1
Pod version 1.5.1

public protocol TestProtocol {
    var didWriteValueFor: (_ uuid: CBUUID) -> () { get set }
}

public class Test: TestProtocol {
    public var didWriteValueFor: (_ uuid: CBUUID) -> () = { fffff in }
}

MockGenerates:     

public var didWriteValueFor: (_ uuid: CBUUID) -> () {
        get {
            return DefaultValueRegistry.defaultValue(for: ((_ uuid: CBUUID) -> ()).self)
        }
        
        set { }
        
    }

Mock generates with errors

<img width="787" alt="Screenshot 2021-06-28 at 17 21 44" src="https://user-images.githubusercontent.com/5227456/123652848-52dccb00-d835-11eb-92d8-cdd9f718adc5.png">

MatyasKriz:
Thanks for the bug report, it seems like the variable type needs to be stripped of the parameter names in this case. Does it work with just `(CBUUID) -> ()`?

evgzor:
I used typealias at the end. No it does not work for me.

MatyasKriz:
I see, thanks for trying it out, must be a different way then.

sk409:
@MatyasKriz
I think the solution is to remove the argument (like ` (CBUUID) -> ()` )
Can I work on this issue to implement solution like above?

MatyasKriz:
Hey, @sk409. Yeah, that sounds like the right solution to this. 🙂 Please go ahead.

PR #433, opened 2022-08-28
Remove closure argument name.

sk409:
# Summary
I removed closure argument name to fix this issue(#394).

# Problem
Closure argument name is not allowed to be used as `Type`.

This problem caused compile error like this:

```swift
DefaultValueRegistry.defaultValue(for: ((_ uuid: CBUUID) -> ()).self)
```

- **Expected ',' separator insert ','**
- **Expected type before '->'**

# Solution
My solution is to remove argument name like this:

```swift
DefaultValueRegistry.defaultValue(for: ((CBUUID) -> ()).self)
```

Thank you for your confirmation.

### DIFF
--- a/Tests/Swift/Source/TestedClass.swift
+++ b/Tests/Swift/Source/TestedClass.swift
@@ -38,9 +38,15 @@ class TestedClass {
 
     var closureProperty: ((String) -> Int) -> () = { i in }
 
-    var closureWithArgumentNameProperty:  (_ i_iI0: (_ j_jJ0: String) -> Int) -> () = { i in }
+    var closureWithArgumentNameProperty: (_ i: (_ j: String) -> Int) -> () = { i in }
 
     var closureWithArgumentNameMultilineProperty: (
+        _ i: (_ j: Int) -> String
+    ) -> () = { i in }
+
+    var closureWithComplexArgumentNameProperty:  (_ i_iI0: (_ j_jJ0: String) -> Int) -> () = { i in }
+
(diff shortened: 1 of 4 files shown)


Tests/Swift/Source/TestedClass.swift
```diff
@@ -36,6 +36,17 @@ class TestedClass {
 
     lazy var optionalProperty: Int? = 0
 
+    var closureProperty: ((String) -> Int) -> () = { i in }
+
+    var closureWithArgumentNameProperty:  (_ i_iI0: (_ j_jJ0: String) -> Int) -> () = { i in }
```
MatyasKriz, line 41: Would you please add a property with some normal-looking parameter names as well? 😄

MatyasKriz:
Looks good, thanks for adding the tests as well. 🙂

Tests/Swift/Source/TestedClass.swift
sk409: @MatyasKriz
Thank you for your review.
I added some normal-looking parameter names.
Is this([ae9fa96](https://github.com/Brightify/Cuckoo/pull/433/commits/ae9fa966fc46426858c52c6afc915387c9976aab)) what you were talking about? 🙂

MatyasKriz: Yes, that's exactly what I meant. 🙂 I'm looking forward to merging this, thanks a lot for the PR!

MatyasKriz:
Merged! 🙂 I'll release a new version right away, as this seems to me like it's better fixed as soon as possible.

Issue #277, opened 2019-05-09
After / delay with verify


Ufosek:
Hello,

Is there an option in Cuckoo to use `after` or other `delay` option with matcher? Something similar to: [VerificationAfterDelay](https://static.javadoc.io/org.mockito/mockito-core/2.6.5/org/mockito/verification/VerificationAfterDelay.html)

For now I combined it with `expectations` but I am not sure whether it's the best way.

```
func test() {
    let exp = expectation(description: "testShow")
    exp.expectedFulfillmentCount = 1

    delay(1.0) { // async after
        verify(view).show()
        exp.fulfill() // always fulfill
    }

    wait(for: [exp], timeout: 2.0)
}
```

USE CASE:
I wanted to test whether some methods were invoked in `Presenter` which used `Timer`. I could of course mock this `Timer` as it is usually done with API requests but for such simple scenario I think delay in verify would be better.

Regards

MatyasKriz:
Hey there, at the moment Cuckoo does not support these async matchers.

It's something that I can imagine is used often, as others have already asked. However, at the moment our greatest priority is generics support.

Though you can implement this functionality if you're up for it and submit a PR. It will probably need to be a part of Cuckoo sooner or later.

sk409:
@MatyasKriz 
I submitted a PR(#434).
This PR is a prototype, so I haven't written any comments, tests, etc.
If this PR is acceptable, I will write comments, tests and explanation to README.md.
If there are any improvements, please point them out.

andreasziegl:
@sk409  @MatyasKriz  any news on this topic? :)

MatyasKriz:
Yeah, some time ago I reviewed the PR but didn't publish the comments. When I got to the PR again, I was unsure if the comments were valid, so I'll have to do another review round.

PR #434, opened 2022-09-03
Introduce timeout and after to support delayed verification.

sk409:
I introduced these two global functions to support delayed verification discussed in this issue(#277).

```swift
// timeout waits until CallMatcher::matches returns true and exits on success.
public func timeout(_ timeoutDuration: TimeInterval, waitingDuration: TimeInterval = 0.01) -> ContinuationWithTimeout
```

```swift
// after waits until delayDuration elapsed and exits when CallMatcher::matches never return true.
public func after(_ delayDuration: TimeInterval, waitingDuration: TimeInterval = 0.01) -> ContinueationAfterDelay
```

`ContinuationWithTimeout` and `ContinueationAfterDelay` conform to `Continuation` protocol.

`Continuation` determines whether CallMatcher::matches should be called again, waits, etc.

like this:

```swift
// MockManager
public func verify<IN, OUT>(_ method: String, callMatcher: CallMatcher, parameterMatchers: [ParameterMatcher<IN>], continuation: Continuation, sourceLocation: SourceLocation) -> __DoNotUse<IN, OUT> {
        ...
        while continuation.check() {
            calls = []
            indexesToRemove = []
            for (i, stubCall) in stubCalls.enumerated() {
                if let stubCall = stubCall as? ConcreteStubCall<IN> , (parameterMatchers.reduce(stubCall.method == method) { $0 && $1.matches(stubCall.parameters) }) {
                    calls.append(stubCall)
                    indexesToRemove.append(i)
                }
            }
            matches = callMatcher.matches(calls)
            if !matches && !callMatcher.canRecoverFromFailure(calls) {
                break
            }
            if matches && continuation.exitOnSuccess {
                break
            }
            continuation.wait()
        }
        ...
        return __DoNotUse()
    }
```

Users can simply write test of fire-and-forget asynchronous function and asynchronous function with completion closure that does not require argument verification.

```swift
stub(mockTestClass) { stub in
    when(stub.f()).thenDoNothing()
}
// someMethod is expected to call f three times within three seconds.
mockTestClass.someMethod()
verify(mockTestClass, timeout(3).times(3)).f()
```

But I think that if users want to verify completion closure argument, They should use expectation.

```swift
stub(mockTestClass) { stub in
    when(stub.f()).thenDoNothing()
}
let expectation = XCTestExpectation(description: #function)
mockTestClass.someMethod() { i in
    XCTAssertEqual(i, 1)
    expectation.fulfill()
}
wait(for: [expectation], timeout: 3)
verify(mockTestClass, times(3)).f()
```

If users write below code, XCTAssertEqual may no be called.

```swift
stub(mockTestClass) { stub in
    when(stub.f()).thenDoNothing()
}
mockTestClass.someMethod() { i in
    XCTAssertEqual(i, 1)
}
verify(mockTestClass, timeout(3).times(3)).f()
```

This PR is a prototype.
If this PR is acceptable, I will write comments, tests and explanation to README.md.

If there are any improvements, please point them out.

### DIFF
--- a/Source/Matching/CallMatcher.swift
+++ b/Source/Matching/CallMatcher.swift
@@ -11,33 +11,44 @@ public struct CallMatcher {
     public let name: String
     
     private let matchesFunction: ([StubCall]) -> Bool
+    private let canRecoverFromFailureFunction: ([StubCall]) -> Bool
     
-    public init(name: String, matchesFunction: @escaping ([StubCall]) -> Bool) {
+    public init(name: String, matchesFunction: @escaping ([StubCall]) -> Bool, canRecoverFromFailureFunction: @escaping ([StubCall]) -> Bool) {
         self.name = name
         self.matchesFunction = matchesFunction
+        self.canRecoverFromFailureFunction = canRecoverFromFailureFunction
     }
-    
-    public init(name: String, numberOfExpectedCalls: Int, compareCallsFunction: @escaping (_ expected: Int, _ actual: Int) -> Bool) {
-        self.init(name: name) {
-            return compareCallsFunction(numberOfExpectedCalls, $0.count)
+
+    public init(
+        name: String,
+        numberOfExpectedCalls: Int,
+        compareCallsFunction: @escaping (_ expected: Int, _ actual: Int) -> Bool,
+        canRecoverFromFailureFunction: @escaping (_ expected: Int, _ actual: Int) -> Bool
+    ) {
+        self.init(name: name, matchesFunction: { compareCallsFunction(numberOfExpectedCalls, $0.count) }) {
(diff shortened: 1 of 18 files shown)


Source/Matching/CallMatcher.swift
```diff
@@ -11,33 +11,44 @@ public struct CallMatcher {
     public let name: String
     
     private let matchesFunction: ([StubCall]) -> Bool
+    private let canRecoverFromFailureFunction: ([StubCall]) -> Bool
```
MatyasKriz, line 14: Why is this necessary?

MatyasKriz:
I like the API this provides, but I need some more context on the implementation before we go ahead with it.

PR #435, opened 2022-09-04
Support NSObjectProtocol inheritance for protocol.

### LINKED ISSUE #425: Support NSObjectProtocol inheritance for protocol
When having a protocol define inheritance from `NSObjectProtocol` like this:

```swift
protocol TestProtocol: NSObjectProtocol { … }
```

Cuckoo generates:
```swift
class MockTestProtocol: TestProtocol, Cuckoo.ProtocolMock { … }
```

which produces the error:
> cannot declare conformance to 'NSObjectProtocol' in Swift; 'MockTestProtocol' should inherit 'NSObject' instead

So instead, whenever we try to generate a Mock from a protocol that inherits from NSObjectProtocol, a fix would be to let it inherit from `NSObject`:

```swift
class MockTestProtocol: NSObject, TestProtocol, Cuckoo.ProtocolMock { … }
```

Motivation: When interfacing with Obj-C code there are a lot of NSObjects involved. So this would be super handy to have.

Has someone a hint how this could be implemented? I'm could open a PR but as the generator code base is quite complex, I would need some hints if this is easily doable and potentially where modifications are needed.

sk409:
I introduced support for NSObjectProtocol inheritance to fix #425.

If protocol inherits from NSObjectProtocol, Generated Mocks inherits from NSObject.

I used the logic suggested by @MatyasKriz with a slight modification to support Protocol Composition. 
```swift
// To support Protocol Composition like NSObject & SomeProtocol, I used following logic.
// Please see test protocol that I added.
let collapsedInheritedTypesName = protocolDeclaration.inheritedTypes.map({ $0.name.components(separatedBy: "&") }).joined().map({ $0.trimmingCharacters(in: .whitespaces) })
if collapsedInheritedTypesName.contains(where: { $0 == "NSObjectProtocol" }) {
    return true
}
```

Thanks.

### DIFF
--- a/Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
+++ b/Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
@@ -38,7 +38,7 @@ public struct GenerateMocksCommand: CommandProtocol {
         }
         let inputFiles = inputPathValues.map { File(path: $0) }.compactMap { $0 }
         let tokens = inputFiles.map { Tokenizer(sourceFile: $0, debugMode: options.debugMode).tokenize() }
-        let tokensWithInheritance = options.noInheritance ? tokens : mergeInheritance(tokens)
+        let tokensWithInheritance = options.noInheritance ? tokens : inheritNSObject(mergeInheritance(tokens))
 
         // filter classes/protocols based on the settings passed to the generator
         var typeFilters = [] as [(Token) -> Bool]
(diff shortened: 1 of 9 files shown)


Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
```diff
@@ -86,6 +86,39 @@ public struct GenerateMocksCommand: CommandProtocol {
         return filesRepresentation.compactMap { $0.mergeInheritance(with: filesRepresentation) }
     }
 
+    private func inheritNSObject(_ filesRepresentation: [FileRepresentation]) -> [FileRepresentation] {
+        func containsRecursively(name: String) -> Bool {
+            if let protocolDeclaration = protocolDeclarationDictionary[name] {
+                let collapsedInheritedTypesName = protocolDeclaration.inheritedTypes.map({ $0.name.components(separatedBy: "&") }).joined().map({ $0.trimmingCharacters(in: .whitespaces) })
```
MatyasKriz, line 92: Please break these method calls to their own lines, this line is way too long to be easy to read.

MatyasKriz:
Thanks for the PR, @sk409. 🙂 I don't really like the new extension methods, as I think this violates the SRP, but if I understand it correctly, you went by the `mergeInheritances` example. It's not that big of a deal and can be refactored later.

I'll merge this as soon as the single nitpick is fixed.

Generator/Source/cuckoo_generator/GenerateMocksCommand.swift
sk409: I breaked these method calls to their own lines.

sk409:
@MatyasKriz 
Thank you for your review.
I agree with you about SRP.
As you said, I implemented `inheritNSObject` with reference to `mergeInheritance`.
I will not use extension methods next time.

MatyasKriz:
Hey, @sk409! I've merged the PR, sorry for the wait. 🙂 Though this is new functionality and I don't feel like it justifies a minor release on its own, so I'll wait a little longer on that.

Issue #452, opened 2023-02-06
Unable to mock functions with certain keywords as argument labels


sethdeckard:
Generate a mock for the following protocol:

```swift
protocol TestProtocol {
    func test(for: String)
}
```

This produces code that will not compile because the keyword when used in the body of the method is not escaped:
```swift
     func test(for: String)  {
        
    return cuckoo_manager.call(
    """
    test(for: String)
    """,
            parameters: (for), // error here
            escapingParameters: (for), // error here
            superclassCall:
                
                Cuckoo.MockManager.crashOnProtocolSuperclassCall()
                ,
            defaultCall: __defaultImplStub!.test(for: for)) // error here (and a few other places)
        
    }
```

PR https://github.com/Brightify/Cuckoo/pull/412 seems to attempt to address this, but only for the function name itself. Which kind of seems odd given that using `for` as an argument label seems somewhat common compared to using it as a function name, which seems like more of an edge case. In fact, Foundation actually has several functions that use `for:` as an argument label.
 
The release notes mention this as being fixed but I think that should have been limited to function names (probably due to commit title being too general):

> Reserved keywords are now escaped. (thank you, @sk409!)

Cuckoo version: 1.9.1

sethdeckard:
It looks like I was able to patch this on my side, by mostly changing the templates, but also some of the filters. I'll submit a PR in the am. 😄

PR #454, opened 2023-02-09
Fix Generator to handle reserved keywords

sethdeckard:
Allows for reserved keywords to be used as argument labels / parameters
in mocked types.

Fixes: https://github.com/Brightify/Cuckoo/issues/452

- Moves escaping util function to Utils and makes it internal since it
  now needs to be used outside of `Generator`.
- Uses escaping function in areas affected by the bug.
- Renames set of reserved names to reflect the larger scope (no longer
  used for just function names).
- Some minor clean-up to style and formatting on areas touched.

Example of the problem:

Generate a mock for a protocol with the function:

```
func escape(for: String) -> String
```

This would result in `for` as a function parameter used within the body
of the function for the mock and verification and stubbing proxies,
which would not compile. This change escapes keywords like this so the
generated code will compile.

See `TestedProtocol.withReservedKeywords` for a test example. An example
was added to `TestedClass` as well, but the real motivation for this is
protocols since conforming types can and often do use a different parameter
name so the argument label like `for` would never be used within the
body of the function.

### DIFF
--- a/Generator/Source/Internal/Generator.swift
+++ b/Generator/Source/Internal/Generator.swift
@@ -2,18 +2,6 @@ import Foundation
 import Stencil
 
 public struct Generator {
-    
-    private static let reservedKeywordsNotAllowedAsMethodName: Set = [
-        // Keywords used in declarations:
-        "associatedtype", "class", "deinit", "enum", "extension", "fileprivate", "func", "import", "init", "inout", "internal", "let", "operator", "private", "precedencegroup", "protocol", "public", "rethrows", "static", "struct", "subscript", "typealias", "var",
-        // Keywords used in statements:
-        "break", "case", "catch", "continue", "default", "defer", "do", "else", "fallthrough", "for", "guard", "if", "in", "repeat", "return", "throw", "switch", "where", "while",
-        // Keywords used in expressions and types:
-        "Any", "as", "catch", "false", "is", "nil", "rethrows", "self", "super", "throw", "throws", "true", "try",
-        // Keywords used in patterns:
-        "_",
-    ]
-
     private let declarations: [Token]
 
     public init(file: FileRepresentation) {
@@ -56,10 +44,10 @@ public struct Generator {
             guard let parameters = value as? [MethodParameter] else { return value }
             return self.closeNestedClosure(for: parameters)
         }
-        
+
         ext.registerFilter("escapeReservedKeywords") { (value: Any?) in
             guard let name = value as? String else { return value }
-            return self.escapeReservedKeywords(for: name)
+            return escapeReservedKeywords(for: name)
         }
 
         ext.registerFilter("removeClosureArgumentNames") { (value: Any?) in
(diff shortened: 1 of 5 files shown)


Generator/Source/Internal/Utils.swift
```diff
@@ -48,9 +48,29 @@ extension Sequence {
     }
 }
 
+/// Utility function for escaping reserved keywords for a symbol name.
+internal func escapeReservedKeywords(for name: String) -> String {
```
MatyasKriz, line 52: This function would be better positioned right above the global constant that it references since it's also the only place said constant is used.

Generator/Source/Internal/Generator.swift
```diff
@@ -110,7 +98,13 @@ public struct Generator {
         guard parameters.isEmpty == false else { return "let matchers: [Cuckoo.ParameterMatcher<Void>] = []" }
 
         let tupleType = parameters.map { $0.typeWithoutAttributes }.joined(separator: ", ")
-        let matchers = parameters.enumerated().map { "wrap(matchable: \($1.name)) { $0\(parameters.count > 1 ? ".\($0)" : "") }" }.joined(separator: ", ")
+
+        let matchers = parameters.enumerated().map { (index, parameter) in
```
MatyasKriz, line 102: A nitpick, but please omit the parentheses around `index, parameter`.

MatyasKriz:
Hey there, thanks for the PR! Though I wouldn't call the previous PR "an attempt" as it did fix what it set out to do and defined all the reserved keywords, which doesn't come free either. This PR extends that further, which is how it should work. 🙂 Either way, I'm looking forward to merging this after some little changes, then we can release it as a patch version.

Generator/Source/Internal/Generator.swift
sethdeckard: If you want contributors to use a specific style, then you should consider adding something like SwiftLint to the project so these style changes can be made before the PR is submitted or caught with a GH check during creation.

With linting enforced as GH check, you'd avoid back and forth like this completely and make it easier for folks to contribute.

Generator/Source/Internal/Utils.swift
sethdeckard: Feel free to refactor and reformat as you wish.

sethdeckard:
You're welcome @MatyasKriz.

I was initially going off the release notes and commit title which only said "Escape reserved keywords". So I spent a bit of time trying to understand why it wasn't working correctly, thinking there was a problem on my side.

I eventually looked at the source and realized it was only handling method names. I was a bit frustrated at the time. 😅

sethdeckard:
@MatyasKriz I pushed again to handle the merge conflict with f0f7739. Since I had to push again, I went ahead and made the minor style suggestions as well.

MatyasKriz:
> @MatyasKriz I pushed again to handle the merge conflict with [f0f7739](https://github.com/Brightify/Cuckoo/commit/f0f77397a58b049ebdcf3017a3daf6bea9fa25b7). Since I had to push again, I went ahead and made the minor style suggestions as well.

Thanks. 🙂 I just noticed that this PR doesn't include tests showcasing this new feature and also making sure that it's compilable and not broken by future changes. Will you add some examples to `TestedClass.swift` perhaps?

sethdeckard:
@MatyasKriz Just to be clear this is a bug fix, not a new feature.

Earlier you mentioned that you would merge the PR after the style changes were performed and now you're asking for examples to be added `TestedClass.swift`.

Is there anything else?

MatyasKriz:
> @MatyasKriz Just to be clear this is a bug fix, not a new feature.

I understand that you were frustrated, but I disagree with you in that regard, it's an extension to an existing feature. 🙂

Since this PR adds something new, do you think it's unreasonable to want to add tests so that it's safe from future changes?

I promise to take your advice to heart and add Swiftlint and a PR checklist as soon as I can. 😁

sethdeckard:
@MatyasKriz I'm onboard with tests, I'm only asking to make sure there is nothing else you require (wanting to reduce some back and forth).

MatyasKriz:
> @MatyasKriz I'm onboard with tests, I'm only asking to make sure there is nothing else you require (wanting to reduce some back and forth).

Great! 🙂 There's nothing else that I can think of.

MatyasKriz:
> @MatyasKriz This is ready with test examples added.

Great! I'll merge this and release it. 🙂 Thanks, @sethdeckard!

PR #458, opened 2023-03-28
Fix property wrapper crash

### LINKED ISSUE #341: Mock generation fails with certain property wrappers
## Expected Behavior
The run script successfully generates all mocks.

## Current Behavior
The run script crashes with the message `Fatal error: String index is out of bounds`

## Steps to Reproduce
1. Create two files:
```swift
class UsefulClass {
    @SomeWrapper(useValue: "foo")
    private var myString
    
}
```
```swift
@propertyWrapper
struct SomeWrapper {
    var wrappedValue: String

    init(useValue: String) {
        self.wrappedValue = useValue
    }
}
```

Compile test sources with the following script (Try to mock `UsefulClass`):
```
# Define output file. Change "${PROJECT_DIR}/${PROJECT_NAME}Tests" to your test's root source folder, if it's not the default name.
OUTPUT_FILE="${PROJECT_DIR}/${PROJECT_NAME}Tests/GeneratedMocks.swift"
echo "Generated Mocks File = ${OUTPUT_FILE}"

# Define input directory. Change "${PROJECT_DIR}/${PROJECT_NAME}" to your project's root source folder, if it's not the default name.
INPUT_DIR="${PROJECT_DIR}/${PROJECT_NAME}"
echo "Mocks Input Directory = ${INPUT_DIR}"

# Generate mock files, include as many input files as you'd like to create mocks for.
"${PODS_ROOT}/Cuckoo/run" generate --testable "${PROJECT_NAME}" \
--output "${OUTPUT_FILE}" \
"${INPUT_DIR}/UsefulClass.swift"
```
The script will produce the error.

The error can be **avoided** if you give additional type information:

```swift
class UsefulClass {
    @SomeWrapper(useValue: "foo")
    private var myString: String
    
}
```

Notice that here I give the variable `myString` an explicit type. This should not be required however, as the Swift compiler infers this type. With the explicit type, the script generates the mocks without any issue.

kabiroberai:
This PR doesn't endeavour to enable type inference for property wrappers without type annotations, but it does allow them to exist in the user's source without causing Cuckoo to crash. This sort of pattern is super common in SwiftUI views, for example, which often have declarations like `@Environment(\.colorScheme) var colorScheme`.

Fixes #341.

### DIFF
--- a/Generator/Source/Internal/TypeGuesser.swift
+++ b/Generator/Source/Internal/TypeGuesser.swift
@@ -3,6 +3,8 @@ import Foundation
 struct TypeGuesser {
     static func guessType(from value: String) -> String? {
         let value = value.trimmed
+        guard !value.isEmpty else { return nil }
+
         let casting = checkCasting(from: value)
         guard casting == nil else { return casting }
 
--- a/Tests/Swift/Source/PropertyWrappers.swift
+++ b/Tests/Swift/Source/PropertyWrappers.swift
@@ -5,6 +5,9 @@
 //  Created by Kabir Oberai on 2023-03-28.
 //
 
+// wrappers without annotations aren't supported but their
+// existence shouldn't cause the generator to crash
+
 @propertyWrapper
 struct DoublingWrapper {
     private var backing: Int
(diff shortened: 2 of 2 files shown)


MatyasKriz:
Looks good, thanks for the PR and test. 🙂

MatyasKriz:
Merged. 🙂

PR #459, opened 2023-03-28
Add SPM package plugin

### LINKED ISSUE #438: [Suggestion] Swift Package Manager plugin
Hello,

First of all thanks for all the good work, guys. We've been using Cuckoo as a cocoapod for a long time now and we're currently in a process of trasitioning to use SPM as our only package manager.

Is there currenty any option to run Cuckoo generator inside a SPM package? I've read the Readme, but the  instructions for SPM installation describe steps needed to use the generator inside a XCode project, not a SPM package.

The correct way to allow for generator usage inside a package would be to utilize SPM Plugins which are now supported in Xcode 14. A good example of that is [SwiftGenPlugin](https://github.com/SwiftGen/SwiftGenPlugin). I thinkt the first step towards allowing the community to build such a plugin is to expose an [artifact bundle](https://theswiftdev.com/introduction-to-spm-artifact-bundles/) through github. 

Is that currently possible for Cuckoo?

Thanks.

kabiroberai:
Closes #438.

Using the plugin is simple: just add `plugins: [.plugin(name: "CuckooPlugin", package: "Cuckoo")]` to the corresponding `testTarget`. The plugin currently passes all modules that are _direct_ dependencies of the test target to the generator.

Since there's no API for directly configuring plugins from Package.swift yet, the plugin currently looks for a `cuckoo.json` file in the package root dir, which can currently be used to supply additional command-line options. IMO using json makes sense for now because it's one of the easiest formats to parse in Swift. We can always switch to something more expressive like YAML in the future, since the config file is versioned and YAML is designed to be backwards-compatible with JSON.

### DIFF
--- a/Generator/Plugin/Plugin.swift
+++ b/Generator/Plugin/Plugin.swift
@@ -1,0 +1,44 @@
+import PackagePlugin
+
+@main struct CuckooPlugin: BuildToolPlugin {
+    func createBuildCommands(context: PluginContext, target: Target) async throws -> [PackagePlugin.Command] {
+        let dependencies: [SourceModuleTarget] = target
+            .dependencies
+            .flatMap { dependency in
+                switch dependency {
+                case .product(let product):
+                    return product.targets
+                case .target(let target):
+                    return [target]
+                @unknown default:
+                    return []
+                }
+            }
+            .compactMap { $0 as? SourceModuleTarget }
+            .filter { $0.kind == .generic && $0.moduleName != "Cuckoo" }
+
+        let testableModules = dependencies
+            .map(\.moduleName)
+            .joined(separator: ",")
+
+        let sources = dependencies
+            .flatMap(\.sourceFiles)
+            .filter { $0.type == .source }
+            .map(\.path)
+
+        let output = context.pluginWorkDirectory.appending("GeneratedMocks.swift")
+
+        return [.buildCommand(
+            displayName: "Run CuckooGenerator",
+            executable: try context.tool(named: "CuckooGenerator").path,
+            arguments: [
+                "generate",
+                "--output", output,
+                "--testable", testableModules,
+                "--regex", "Mockable.*"
+            ] + sources,
+            inputFiles: sources,
+            outputFiles: [output]
+        )]
+    }
+}
--- a/Generator/Plugin/Plugin.swift
+++ b/Generator/Plugin/Plugin.swift
@@ -3,7 +3,7 @@ import Foundation
 
 @main struct CuckooPlugin: BuildToolPlugin {
     func createBuildCommands(context: PluginContext, target: Target) async throws -> [PackagePlugin.Command] {
-        let configPath = context.package.directory.appending("cuckoo-config.json")
+        let configPath = context.package.directory.appending("cuckoo.json")
         let config = try await ConfigFile.decode(from: configPath)
 
         let dependencies: [SourceModuleTarget] = target
--- a/Generator/Plugin/Plugin.swift
+++ b/Generator/Plugin/Plugin.swift
@@ -1,7 +1,11 @@
 import PackagePlugin
+import Foundation
 
 @main struct CuckooPlugin: BuildToolPlugin {
     func createBuildCommands(context: PluginContext, target: Target) async throws -> [PackagePlugin.Command] {
+        let configPath = context.package.directory.appending("cuckoo-config.json")
+        let config = try await ConfigFile.decode(from: configPath)
+
         let dependencies: [SourceModuleTarget] = target
             .dependencies
             .flatMap { dependency in
@@ -34,11 +38,40 @@ import PackagePlugin
             arguments: [
                 "generate",
                 "--output", output,
-                "--testable", testableModules,
-                "--regex", "Mockable.*"
-            ] + sources,
-            inputFiles: sources,
+                "--testable", testableModules
+            ] + config.options + sources,
+            inputFiles: [configPath] + sources,
             outputFiles: [output]
         )]
+    }
+}
+
+struct ConfigFile: Codable {
+    enum Version: Int, Codable {
+        case v1 = 1
+    }
+
+    var version: Version = .v1
+    var options: [String] = []
+
+    static func decode(from path: Path) async throws -> Self {
+        guard let data = try? await path.contents() else { return .init() }
+        return try JSONDecoder().decode(self, from: data)
+    }
+}
+
+extension Path {
+    func contents() async throws -> Data {
+        let url: URL
+        if #available(macOS 13, *) {
+            url = URL(filePath: string)
+        } else {
+            url = URL(fileURLWithPath: string)
+        }
+        if #available(macOS 12, *) {
+            return try await url.resourceBytes.reduce(into: Data()) { $0.append($1) }
+        } else {
+            return try Data(contentsOf: url)
+        }
     }
 }
--- a/Package@swift-5.7.swift
+++ b/Package@swift-5.7.swift
@@ -1,0 +1,52 @@
+// swift-tools-version:5.7
+
+import PackageDescription
+
+let package = Package(
+    name: "Cuckoo",
+    platforms: [.macOS("12.0")],
+    products: [
+        .library(
+            name: "Cuckoo",
+            targets: ["Cuckoo"]
+        ),
+        .plugin(
+            name: "CuckooPlugin",
+            targets: ["CuckooPlugin"]
+        ),
+    ],
+    dependencies: [
+        .package(url: "https://github.com/jpsim/SourceKitten.git", .upToNextMinor(from: "0.34.1")),
+        .package(url: "https://github.com/nvzqz/FileKit.git", branch: "develop"),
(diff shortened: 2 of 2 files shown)


Package@swift-5.7.swift
```diff
@@ -0,0 +1,52 @@
...
+            name: "Cuckoo",
+            targets: ["Cuckoo"]
+        ),
+        .plugin(
+            name: "CuckooPlugin",
+            targets: ["CuckooPlugin"]
+        ),
+    ],
+    dependencies: [
+        .package(url: "https://github.com/jpsim/SourceKitten.git", .upToNextMinor(from: "0.34.1")),
```
kabiroberai, line 19: I had to update SourceKitten to a version that's compatible with package plugins (now uses in-proc sourcekit to avoid XPC and pacify sandbox restrictions). This does mean that I had to bump the macOS target to 12.0 though. It might make sense to update the SourceKitten + macOS versions in the Tuist config too.

Package@swift-5.7.swift
```diff
@@ -0,0 +1,52 @@
+// swift-tools-version:5.7
```
MatyasKriz, line 1: Do you think it might make sense to replace the current `Package.swift` instead?

Generator/Plugin/Plugin.swift
```diff
@@ -0,0 +1,77 @@
+import PackagePlugin
+import Foundation
+
+@main struct CuckooPlugin: BuildToolPlugin {
+    func createBuildCommands(context: PluginContext, target: Target) async throws -> [PackagePlugin.Command] {
+        let configPath = context.package.directory.appending("cuckoo.json")
```
MatyasKriz, line 6: Do you think it might be possible to go the way SwiftFormat goes and just use a plain text file with CLI arguments that is then passed into the CLI? I'd like to keep the API consistent before SwiftSyntax is ready, after that we can tackle the beast that is converting CLI options into `CuckooFile` of type TOML perhaps.

To be clear SwiftFormat configuration file is simply a plain text file that contains CLI arguments on their own lines (or ending with `\` if they're a continuation), sort of like this:

```
--testable Cuckoo
--exclude ExcludedTestClass,ExcludedProtocol
--output "./Tests/Swift/Generated/GeneratedMocks.swift"
--glob
"./Tests/Swift/Source/*.swift"
```

Package@swift-5.7.swift
MatyasKriz: I'm not sure how I feel about bumping the version to macOS 12 to be honest, it's quite a jump.

Generator/Plugin/Plugin.swift
kabiroberai: I considered that too but I found a few issues:

1. Parsing escape sequences is _harder_ in a plaintext file since we'd have to do it manually, whereas we get it for free with JSON.
2. We might want to add more options to the current file format besides just adding extra CLI options: for example the ability to specify whether we do shallow or deep traversal of dependencies. This would allow users to make use of the config file while still retaining (and customizing) the automatic analysis that SPM plugins can offer.

Overall on a spectrum of configurability IMO plaintext is on one end, JSON is somewhere in the middle, and a YAML/TOML file would be the endgame. Especially since JSON is simpler to parse I think it's better if we stick with that than take a step back to plaintext. That being said, the final decision is up to you ofc.

Package@swift-5.7.swift
kabiroberai: yeah I wish we could specify separate build-time and runtime versions. One workaround could be to make the generator a separate package, which would also answer your question about the separate Package.swift file. What do you think about taking that approach instead?

Generator/Plugin/Plugin.swift
```diff
@@ -0,0 +1,77 @@
...
+
+extension Path {
+    func contents() async throws -> Data {
+        let url: URL
+        if #available(macOS 13, *) {
+            url = URL(filePath: string)
+        } else {
+            url = URL(fileURLWithPath: string)
+        }
+        if #available(macOS 12, *) {
```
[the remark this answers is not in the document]
kabiroberai: Ah good point, plugins require Swift 5.7 => macOS 12 anyway so we can remove the availability check regardless of the overall deployment target.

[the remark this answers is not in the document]
kabiroberai: oh I see why I added this check — seems like Xcode doesn't realize that the deployment target is macOS 12, and throws a compiler error if the availability check is missing. Probably an SPM/Xcode bug 🙃

Generator/Plugin/Plugin.swift
MatyasKriz: I'm not sure I understand the first point. The current CLI-only solution supports escape sequences?

As for the second point, that's what I'd like to avoid. The configuration must be consistent whether you use the SPM plugin or download the binary. The SPM plugin should be the simplest it can be, not adding "premium" features. 😄 

Another problem this brings is that AFAIK JSON isn't compatible with TOML, so users who want to use the SPM plugin will first have to migrate to JSON and then again to TOML when that feature is ready.

Package@swift-5.7.swift
MatyasKriz: Are there any problems you can think of that might occur if we do separate it? I mean for Tuist it's already separated and it's currently not included as a dependency for anything other than CocoaPods IIRC.

Generator/Plugin/Plugin.swift
```diff
@@ -0,0 +1,77 @@
...
+        let testableModules = dependencies
+            .map(\.moduleName)
+            .joined(separator: ",")
+
+        let sources = dependencies
+            .flatMap(\.sourceFiles)
+            .filter { $0.type == .source }
+            .map(\.path)
+
+        let output = context.pluginWorkDirectory.appending("GeneratedMocks.swift")
```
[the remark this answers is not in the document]
bohdansasko: @kabiroberai @bspinner @MatyasKriz 
what is the status of the PR? We would like to use the functionality in our app 😊

bohdansasko:
@kabiroberai @bspinner @MatyasKriz 
what is the status of the PR? We would like to use the functionality in our app

MatyasKriz:
Hey, I don't think I'll be merging this PR, as I've already started working on TOML configuration file to go with the new release and merging this would require users to migrate twice.

I don't have an ETA if that's on your mind, so go ahead and use the unofficial integration in the meantime if you need it. 🙂

bohdansasko:
> Hey, I don't think I'll be merging this PR, as it's been abandoned and I've already started working on TOML configuration file to go with the new release and merging this would require users to migrate twice.
> 
> I don't have an ETA if that's on your mind, so go ahead and use the unofficial integration in the meantime if you need it. 🙂

May I help you somehow to speed up the feature development?

MatyasKriz:
> > Hey, I don't think I'll be merging this PR, as it's been abandoned and I've already started working on TOML configuration file to go with the new release and merging this would require users to migrate twice.
> 
> > 
> 
> > I don't have an ETA if that's on your mind, so go ahead and use the unofficial integration in the meantime if you need it. 🙂
> 
> 
> 
> May I help you somehow to speed up the feature development?

You're too kind, but I need to let this slowly brew little by little when I have the time. I've managed to add TOML configuration support quite easily thanks to a great library as well as parser and generator concurrency. But I'm not done yet, along with it the README needs to be refurbished using the knowledge I've gathered during implementation. I'd like to include this plugin support as well if possible since we have a configuration file ready for it.

What I fear a little is caching which wasn't done before in this project, so that might be a challenge, though I might release that as 2.1 additionally.

kabiroberai:
Just to chime in, I unfortunately haven't had time to work on this as of late and since @MatyasKriz is working on a TOML config I agree that it doesn't make sense to pursue this PR any further. Now that Swift Macros are stabilized IMO it makes more sense to adopt those instead of a package plugin, but that would be source-breaking so I can see arguments both for and against taking that approach.

@MatyasKriz want me to close this PR? Or should I leave it open for posterity until the new configuration system is implemented?

MatyasKriz:
Please keep it open for now so that I don't forget to credit you, I'll use your implementation minus the JSON as it looks solid to me.

I haven't had much opportunity to dabble with macros yet, mostly because when I tried them, they were SPM-only, which is kinda lame. Is it any better now? Also I can't think of a way to use macros without editing existing source code as this the library's way.

Generator/Plugin/Plugin.swift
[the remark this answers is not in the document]
MatyasKriz: I'm integrating the build plugin with Cuckoo 2.0 and it seems like this isn't possible for 2 reasons:
- AFAIK output paths must be declared in advance through the `outputFiles`
- there's no way to configure build tools, as @kabiroberai stated

I'll have to make sure this deviation from normal functioning is properly documented in README as you advised in a different comment.

MatyasKriz:
Thank you @kabiroberai for the initial implementation! It was insanely helpful during adaptation for the TOML configuration that Cuckoo 2.0 uses.

 I'll close the PR in favor of the SwiftSyntax PR that has the SPM plugin contained within.

PR #460, opened 2023-03-28
Support effectful properties

kabiroberai:
This PR allows Cuckoo to properly handle read-only properties with getters marked as `async` or `throws`, as specified by [SE-0310](https://github.com/apple/swift-evolution/blob/main/proposals/0310-effectful-readonly-properties.md).

Note that I'm not 100% sure I have all the availability annotations correct so lmk if there are any problems with that.

### DIFF
--- a/Generator/Source/Internal/Tokenizer.swift
+++ b/Generator/Source/Internal/Tokenizer.swift
@@ -185,11 +185,19 @@ public struct Tokenizer {
                 guessedType = type
             }
 
+            let effects: InstanceVariable.Effects
+            if let bodyRange = bodyRange {
+                effects = parseEffects(source: source[bodyRange])
+            } else {
+                effects = .init()
+            }
+
             return InstanceVariable(
                 name: name,
                 type: guessedType ?? .type("__UnknownType"),
                 accessibility: accessibility,
                 setterAccessibility: setterAccessibility,
+                effects: effects,
                 range: range!,
                 nameRange: nameRange!,
                 overriding: false,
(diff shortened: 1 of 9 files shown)


Generator/Source/Internal/Tokenizer.swift
```diff
@@ -518,6 +526,34 @@ public struct Tokenizer {
         return ReturnSignature(isAsync: isAsync, throwString: throwString, returnType: returnType ?? WrappableType.type("Void"), whereConstraints: whereConstraints)
     }
 
+    private func parseEffects(source: String) -> InstanceVariable.Effects {
+        var effects = InstanceVariable.Effects()
+
+        let trimmed = source.drop(while: { $0.isWhitespace })
+        guard trimmed.hasPrefix("get") else { return effects }
```
kabiroberai, line 533: This doesn't account for the `attributes? mutation-modifier?` part of the production rule in the official grammar, but I don't think readonly properties can have a mutation modifier or any meaningful attributes — though I might be wrong about the latter. I wonder whether it's time to upgrade to SwiftSyntax-based parsing tbh.

MatyasKriz: I'm already working on moving the generator to SwiftSyntax, but the current generator implementation is very dependent on SourceKitten, so untangling it took some time. I'm still in the process of ironing out the details to get the tests green.

Tests/Swift/Stubbing/StubbingTest.swift
```diff
@@ -124,6 +124,55 @@ class StubbingTest: XCTestCase {
...
+        XCTAssertEqual(resultThrows, 6)
+
+        let resultAsyncThrows = try! await mock.asyncThrowsProperty
+        XCTAssertEqual(resultAsyncThrows, 7)
+
+        verify(mock, times(1)).asyncProperty.get()
+        verify(mock, times(1)).throwsProperty.get()
+        verify(mock, times(1)).asyncThrowsProperty.get()
+
+        enum TestError: Error { case fromThrows, fromAsyncThrows }
```
MatyasKriz, line 153: A local enum is pretty smart for tests, didn't think that would be legal.

Tests/Swift/Stubbing/StubbingTest.swift
```diff
@@ -124,6 +124,55 @@ class StubbingTest: XCTestCase {
...
+            caughtFromThrows = true
+        } catch {}
+        XCTAssert(caughtFromThrows)
+
+        var caughtFromAsyncThrows = false
+        do {
+            _ = try await mock.asyncThrowsProperty
+        } catch TestError.fromAsyncThrows {
+            caughtFromAsyncThrows = true
+        } catch {}
```
MatyasKriz, line 173: What do you think about verifying that it doesn't throw a different error?

MatyasKriz:
Regarding the availability annotations, the one used throughout the project for `async` is
```swift
@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
```

So I think it's fine. 🙂

Tests/Swift/Stubbing/StubbingTest.swift
kabiroberai: If a different error is thrown then `caughtFromAsyncThrows` should remain false and the test will fail. Is that sufficient or am I missing something else here?

Generator/Source/Internal/Tokenizer.swift
kabiroberai: I see, looking forward to that!

Tests/Swift/Stubbing/StubbingTest.swift
MatyasKriz: Yeah, that's fine by me, didn't realize that.

Issue #463, opened 2023-04-11
Compilation fails in Xcode 14.3


nanashiki:
## overview

After updating to Xcode 14.3, Cukoo can no longer be built

## Error

```sh
ld: file not found: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc/libarclite_iphoneos.a
clang: error: linker command failed with exit code 1 (use -v to see invocation)

/Users/xxx/Develop/quipper/ecp-ios/Carthage/Checkouts/Cuckoo/Cuckoo.xcodeproj: warning: The iOS deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 11.0 to 16.4.99. (in target 'Cuckoo-iOS' from project 'Cuckoo')
warning: Run script build phase 'Embed Precompiled Frameworks' will be run during every build because it does not specify any outputs. To address this warning, either add output dependencies to the script phase, or configure it to run in every build by unchecking "Based on dependency analysis" in the script phase. (in target 'Cuckoo-iOS' from project 'Cuckoo')
** ARCHIVE FAILED **
```

## Environment

- Xcode 14.3 (14E222b)
- Apple M1 Max 64GB
- macOS 13.2.1（22D68）

PR #464, opened 2023-04-11
Support Xcode 14.3 (Up OS Requirements)

nanashiki:
## overview

Fix: https://github.com/Brightify/Cuckoo/issues/463

## References

https://stackoverflow.com/questions/75574268/missing-file-libarclite-iphoneos-a-xcode-14-3

### DIFF
--- a/Cuckoo.xcodeproj/project.pbxproj
+++ b/Cuckoo.xcodeproj/project.pbxproj
@@ -2684,7 +2684,7 @@
 					"@executable_path/../Frameworks",
 					"@loader_path/../Frameworks",
 				);
-				MACOSX_DEPLOYMENT_TARGET = 10.9;
+				MACOSX_DEPLOYMENT_TARGET = 10.13;
 				PRODUCT_BUNDLE_IDENTIFIER = org.brightify.Cuckoo;
 				PRODUCT_NAME = Cuckoo;
 				SDKROOT = macosx;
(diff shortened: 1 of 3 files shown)


MatyasKriz:
Hey there, @nanashiki, does the `post_install` script not work for your project? I'd like to avoid bumping up the minimum version unless there's no other way.

nanashiki:
@MatyasKriz Thank you for your reply

I wrote a reference for the solution in CocoaPods, but we are using Carthage in our project and we are in trouble.

MatyasKriz:
> The error content is `ld: file not found: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc/libarclite_iphoneos.a` and is not directly related to DEVELOPMENT_TARGET, but i could only find a way

I see, well, the versions aren't that sky-high, so I don't think it's an unreasonable jump. Could you please edit the deployment targets in the `Project.swift` file? That's the project manifest that the pbxproj file is generated from. Then run `make` again to have it regenerate the pbxproj file.

nanashiki:
@MatyasKriz Edited PlatformType.swift called from Project.swift
[21f132a](https://github.com/Brightify/Cuckoo/pull/464/commits/21f132a55f7b2d1ff1aac0132d7d57fe40eac0fc)

MatyasKriz:
Thank you, I've merged this and will release this momentarily. 🙂

PR #478, opened 2023-10-13
Update README.md

Yanivox:
use build dir source pckages checkout to run cockoo

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,21 +106,14 @@ Note: All paths in the Run script must be absolute. Variable `PROJECT_DIR` autom
 4. On the 'Choose Package Products for Cuckoo' dialog, under 'Add to Target', please ensure you select your Test target as it will not compile on the app target.
 5. Click Add Package.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,22 +106,15 @@ Note: All paths in the Run script must be absolute. Variable `PROJECT_DIR` autom
 4. On the 'Choose Package Products for Cuckoo' dialog, under 'Add to Target', please ensure you select your Test target as it will not compile on the app target.
```
aiKrice, line 106: I would also update this:
3. For the Dependency Rule, Select "Up to Next Major" with `1.10.1`. Click Add Package.
4. On the 'Choose Package Products for Cuckoo' dialog, under 'Add to Target', please ensure you select your Test target as it will not compile on the app target.

MatyasKriz:
Hey, @Yanivox, sorry for not responding sooner. I'll merge this along with @aiKrice's suggestions.

In 2.0 however, SPM plugin will be the preferred way of running the generator through SPM, so this section will change once more.

README.md
MatyasKriz: Oh, just the version. I see.

MatyasKriz:
Merged.

PR #483, opened 2024-01-01
Convert to SwiftSyntax

### LINKED ISSUE #224: Mock classes from third-party libraries
I have a class that accepts an instance of third-party library class in initializer (injection).
But Cuckoo framework does not allow mocking a class that is not in our project. This class is located in third-party pod.

Is there any solution for this?

### LINKED ISSUE #297: Cuckoo doesn't propagate exit status code when generating mocks fails

### LINKED ISSUE #321: Prefix mock- and stubnames

### LINKED ISSUE #348: Add support for input files being supplied from xcode's Build Phase input file list
I've just updated to 1.4.0 to take advantage of the support of XCode Input Files for inputs. It would be great to have support for Input File Lists as well. 

Thanks for keeping on top of these new features

### LINKED ISSUE #397: Import statement is added any time "import something" is found in a comment
Take this class, for example:

`
import Foundation

// some comment has import SOMEBADTHING
public protocol TestProtocol: AnyObject {
  func doSomething() -> Void
}
`

The generated mock ends up having:

`
...
import SOMEBADTHING
...
`

Looks like it is looking for "{import .*}" in source files and just adding that verbatim to the output file, causing compile issues.

### LINKED ISSUE #438: [Suggestion] Swift Package Manager plugin

### LINKED ISSUE #451: Feature Request: add non-testable imports to generator
So we've got a rather complex modular architecture type setup. The unfortunate thing about Cuckoo is that it requires you to point to a particular file path in order to mock something. This isn't such a problem if you own what you're mocking, but if you're in a huge codebase and another team owns it, it makes your mock brittle. 

What we're trying to do:
- Modules can vend their own mocks for their public types

The problem:
Because Cuckoo generates an `@testable import` you get a "module wasn't build for testing" if you try to distribute your mock through a package manager. 

Our current hacky workaround:
- Remove `--testable` from the command line invocation
- Use `sed` to inject our own import haphazardly at the top of the generated file. 

What I'd prefer:
- The generator allows for the same comma-separated list of import statements to be added.

I recognize this may be difficult as the generator determines its imports by looking at imports in the Swift file(s) its mocking, so perhaps this could be `--additionalImports`?

### LINKED ISSUE #455: Use Cuckoo in product target?
Is it possible to use Cuckoo in a product target using SPM?

In our project we usually use Mocks also to render swift ui views previews, or for UI testing.
Would it be possible for the generated mocks to be used in the main target of the application?

I tried something similiar by adding the SPM to the product target and generating the mocks without using the --testable option, and importing them in the product target  but I get the `No such module 'Cuckoo'` error.

### LINKED ISSUE #472: Cuckoo does not support block-type closure
Hi!

I have next protocol with block-type closure:
```swift
protocol AsyncSampleProtocol {
    func async(execute work: @escaping @convention(block) () -> Void)
}

extension AsyncSample : AsyncSampleProtocol {
    func async(execute work: @escaping @convention(block) () -> Void) {
        async(group: nil, qos: .unspecified, flags: [], execute: work)
    }
}
```

When add this files to generator input it fails with:
```bash
Script path: /path/to/generator/CuckooGenerator
xargs: /path/to/generator/cuckoo_generator: terminated with signal 11; aborting
```

### LINKED ISSUE #479: Issue with Cuckoo Generating Mock for a variable's property
**Description **
I'm encountering an issue when using Cuckoo to generate mocks for a property named currentCastedMedia within a protocol. The property is declared with the BehaviorRelay type, and it holds a tuple with several optional and non-optional elements.

**Here's how the property is defined in the protocol:**
`var currentCastedMedia: BehaviorRelay<(streamURL: URL?, title: String, albumName: String, thumbnailURL: URL?, playAtTime: TimeInterval, mediaType: CastMediaType)?> { get }`

**And here's how it's implemented in the class conforming to the protocol:**
`var currentCastedMedia = BehaviorRelay<(streamURL: URL?, title: String, albumName: String, thumbnailURL: URL?, playAtTime: TimeInterval, mediaType: CastMediaType)?>(value: nil)
`

However, when Cuckoo generates mocks for this property, the generated code seems incomplete and not matching the property's declaration.

**Here's a snippet of the generated code:**

Screenshot 2023-11-02 at 9 52 16 PM

Screenshot 2023-11-02 at 9 52 35 PM

Screenshot 2023-11-02 at 9 52 52 PM

Screenshot 2023-11-02 at 9 53 04 PM

The generated code appears to be incomplete and not in alignment with the property's declaration in the protocol and class.

I've verified that the protocol and class implementation are correct, and the issue seems to be related to how Cuckoo handles this specific property.

**Steps Taken:**
1. Verified the correctness of the protocol and class declarations.
2. Manually added the incomplete property and cleared the Generate Cuckoo Mock in Build Phases so it won't change back the manually added fixes.

**Expectation:**
I expect Cuckoo to generate mocks for the currentCastedMedia property in a way that accurately reflects the property's declaration in the protocol and class.

Any guidance or insights into this issue would be greatly appreciated.

### LINKED ISSUE #481: Issue with variable name as `extension`
Hi team I ran into an issue in my class there is a variable named `extension`, when generating the mocks I get this error message 

```
Getter/setter can only be defined for a single variable
Keyword 'extension' cannot be used as an identifier here
If this name is unavoidable, use backticks to escape it
```

The class looks like:

```swift
class PurchasePathInterfaceProxy: NSObject, PurchasePathInterface {
    var `extension`: PurchasePathExtension?

    var isBrowserEnabled: Bool = false

    func fetch(request: PurchasePathRequest, completion: @escaping (PurchasePath?, Error?) -> Void) {
    }
}
```

Screenshot 2023-12-19 at 16 31 35

MatyasKriz:
Lots of breaking changes.

TO DO:
- [x] update README with TOML configuration file syntax
- [x] create beta release
- [x] add all the issues that get fixed when this is merged

Closes #451, closes #438, closes #455, closes #472, closes #479, closes #481, closes #224, closes #348, closes #297, closes #321, closes #397.


MatyasKriz:
> If you need beta tester, feel free to ping me 👍 

Thanks, I'll make sure to do that, I'm now getting README ready for this release since the configuration is done in its own file instead of through CLI options, but right after that I intend to release a beta build for wide use. 🙂

MatyasKriz:
@aiKrice beta.1 is up if you wanna try it out. 🙂

https://github.com/Brightify/Cuckoo/releases/tag/2.0.0-beta.1

aiKrice:
@MatyasKriz I will try it out this week, I have to work on the SPM setup of the project of my cie.
Come back to you if I see any issue :).

aiKrice:
Hello @MatyasKriz I am facing issue when I try to integrate the Beta. 

`Cannot find 'Glibc' in scope` in Path+Extras in XcodeProj package.

MatyasKriz:
> Hello @MatyasKriz I am facing issue when I try to integrate the Beta. 
> 
> 
> 
> `Cannot find 'Glibc' in scope` in Path+Extras in XcodeProj package.

Hey, thanks for trying it out! 🙂 When does this error occur? When you try running the generator?

MatyasKriz:
> Hello @MatyasKriz I am facing issue when I try to integrate the Beta.
> 
> `Cannot find 'Glibc' in scope` in Path+Extras in XcodeProj package.

I've looked at the file you mentioned and based on the condition
```swift
#if os(macOS)
let systemGlob = Darwin.glob
#else
let systemGlob = Glibc.glob
#endif
```

It seems you're not running on macOS. What OS are you running the generator on? It might be necessary to install the Glibc somehow to be able to use it.

MatyasKriz:
Okay, it's been long enough, we'll test this in production then. 😄

PR #510, opened 2024-09-19
Bugfix: fix accessor for typealias

aim2120:
### Bug

If a class had a typealias with an accessor, this value was not used in the mocks generation. This creates build errors if the typealias is used in APIs that require that level of accessor.

```swift
public class MyClass {
  public typealias MyValue = String // <- this typealias would be included in the mock w/o a public accessor
  public func doSomething(with value: MyValue) { }
}
```

### Fix

Add the container's accessibility level to each typealias definition in the template.

### Testing

I've added a new class for testing `ClassWithTypeAlias` that prior to the fix would not create a buildable Swift code.

### DIFF
--- a/Cuckoo.xcodeproj/project.pbxproj
+++ b/Cuckoo.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 52;
+	objectVersion = 55;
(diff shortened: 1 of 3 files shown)


MatyasKriz:
Hey, that's a good catch, not sure why I didn't think of that during implementation. I'll merge it right away. 🙂

PR #515, opened 2024-10-04
Add ability to customize start/end of each generated file

jeffpearce:
This PR adds two properties to the options, to allow per-file customization. For example, you can use this to turn swiftlint on and off, but there are other uses.

```
[modules.MyProject.options]

# You can add prefix lines to the beginning of each file written, and suffix lines to the end
# For example, this config can be used to turn swiftlint off at the start and back on at the end
prefixLines = [
    "// swiftlint:disable all",
]
SuffixLines = [
    "// swiftlint:enable all",
]

```

### DIFF
--- a/Cuckoofile.toml
+++ b/Cuckoofile.toml
@@ -12,10 +12,17 @@ exclude = [
 ]
 # regex = ""
 output = "Tests/Swift/Generated/GeneratedMocks.swift"
+# output = "Tests/Swift/Generated/"
 
 [modules.Cuckoo.options]
 glob = true
 keepDocumentation = false
+filePrefix = [
+    "// swiftlint:disable all",
+]
+fileSuffix = [
+    "// swiftlint:enable all",
+]
 
 [modules.Cuckoo.xcodeproj]
 path = "Generator"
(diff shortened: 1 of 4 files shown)


MatyasKriz:
Hey, thanks for the PR, although I don't believe the generator needs this functionality. I'd like to keep the configuration as simple as can be. Not saying that customization is bad, we've already got nearly infinitely customizable mock names, it's just that these file prefixes/suffixes can be supplemented by a bash line if a project needs it.

What other use-cases have you got in mind for this new feature?

jeffpearce:
Hey, thanks for looking! 

My first approach actually was to run a shell script that uses sed to in add the lines. Unfortunately, there's a race condition under parallel builds, since the script can be called from two different projects, causing the lines to be added twice. I have other ideas to fix it, but if I look at our build as a whole the "simplest" approach is for the generator to be able to output what we need in one step.

As far as other uses, I could envision someone wanting to inject a copyright header into the generated mocks file. Not important to me, but I've worked places where stuff like that was important.

aiKrice:
Agree, more over, yo ucan just exclude swiftlint from these files or from other tools

MatyasKriz:
Sorry, this hasn't gathered any support while open, closing.

PR #516, opened 2024-10-08
Temporarily add `@unchecked Sendable` to every mock.

MatyasKriz:
If you have mocks that make use of `Sendable` or otherwise use Swift concurrency, please try integrating this branch and voting with reactions whether everything works as expected.


MatyasKriz:
Thanks for letting me know, I'll see what I can do. 🙂

MatyasKriz:
Okay, the fix is pushed. I decided to remove the previous implementation, it was too fragile anyway. It's now just hardcoded in the templates you provided. @stzouvaras

PR #519, opened 2024-11-07
Update Version.swift

danibachar:
The hard coded Version should be updated - shouldn't it?
I do wonder if this file should be automatically generated during the cuckoonator build according to the version in the `version` file that is updated no every release?


MatyasKriz:
Turns out it didn't slip my mind, it's part of the process, but it seems like it doesn't work as expected.

PR #527, opened 2024-12-10
Fixed typeErasure template when method.isThrowing

stzouvaras:
typeErasure template was missing the logic to check for method throws while assigning defaultImpl


MatyasKriz:
Looks good, thank you for the PR!

PR #530, opened 2024-12-24
Public imports option for configuration

dylanmaryk:
Due to imports being internal by default from Swift 6, it is sometimes necessary to explicitly make an import `public`. This MR introduces the ability to add public imports to generated files through the Cuckoofile.


MatyasKriz:
Hey, @dylanmaryk, nice contribution! I've skimmed over it and the changes look good. I'll read up more on the scoped imports and merge it as soon as I can. 🙂

MatyasKriz:
Merged!

PR #537, opened 2025-01-28
Update contribute section in README.md to include `make init`

mnaruse:
This PR updates the “Contribute” section in the README.md to include instructions for using `make init` during the initial setup.

## Why this change is necessary

The current README.md does not mention the `make init` command, which is required for the first-time setup. Adding this information helps new contributors set up the project correctly without confusion.

## What this PR does

Adds a step to use `make init` for the initial setup in the “Contribute” section.

## Testing

N/A – This is a documentation-only update.


MatyasKriz:
That's a great addition, thank you. 🙂

PR #541, opened 2025-02-11
Swift 6 strict concurrency checking errors fix

dylanmaryk:
This PR fixes all the errors introduced if using Swift 6's strict concurrency checking. These are not optimal fixes that are reliably safe in terms of preventing races, but they allow Cuckoo to compile successfully with strict concurrency checking enabled.


MatyasKriz:
Oops, I tried updating the branch so I can simply merge it, but it created a merge commit which I cannot accept. 😄 I'll push your changes but I'll have to close this PR.

MatyasKriz:
I also tried these changes (+ Swift 6 in Package.swift) in a local project and it ran without a problem. I'll hold on until a release, but I thank you for these fixes. 🙂

Issue #533, opened 2025-01-09
How to point at correct @escaping @Sendable stub?


chaseklingelzen:
Hi, 

We have the following protocol:

```Swift
public protocol NetworkMonitorProvider {
    ...
    func observeNetworkStatus(_ handler: @escaping @Sendable (_ newPath: NetworkStatusProvider) -> Void)
    ...
}

extension NWPathMonitor: NetworkMonitorProvider {
    public func observeNetworkStatus(
        _ handler: @escaping @Sendable (_ newPath: NetworkStatusProvider) -> Void
    ) {
        pathUpdateHandler = { path in
            handler(path)
        }
    }
}
```

Our generated mock creates the following (note the difference where one actually creates @escaping @Sendable and the other does not):

```Swift
public class NetworkMonitorProviderStub:NetworkMonitorProvider, @unchecked Sendable {
    
    public var path: NetworkStatusProvider {
        get {
            return DefaultValueRegistry.defaultValue(for: (NetworkStatusProvider).self)
        }
    }


    
    public func observeNetworkStatus(_ p0: @escaping @Sendable (_ newPath: NetworkStatusProvider) -> Void) {
        return DefaultValueRegistry.defaultValue(for: (Void).self)
    }
    
    public func start(queue p0: DispatchQueue) {
        return DefaultValueRegistry.defaultValue(for: (Void).self)
    }
    
    public func cancel() {
        return DefaultValueRegistry.defaultValue(for: (Void).self)
    }
}

func observeNetworkStatus<M1: Cuckoo.Matchable>(_ p0: M1) -> Cuckoo.ProtocolStubNoReturnFunction<( (_ newPath: NetworkStatusProvider) -> Void)> where M1.MatchedType ==  (_ newPath: NetworkStatusProvider) -> Void {
            let matchers: [Cuckoo.ParameterMatcher<( (_ newPath: NetworkStatusProvider) -> Void)>] = [wrap(matchable: p0) { $0 }]
            return .init(stub: cuckoo_manager.createStub(for: MockNetworkMonitorProvider.self,
                method: "observeNetworkStatus(_ p0: @escaping @Sendable (_ newPath: NetworkStatusProvider) -> Void)",
                parameterMatchers: matchers
            ))
        }
```

Our unit test is set up like so:

```Swift
stub(monitor) { stub in
      stub.observeNetworkStatus(any()).thenDoNothing()
}
```

The issue is the stub is not pointing at this func:

```Swift
[...]

PR #544, opened 2025-02-26
Fix crash when stubbing function that takes Sendable closure

dylanmaryk:
This PR adds support for correctly mocking and therefore stubbing functions that take a Sendable closure. Previously these functions were not correctly mocked and therefore stubs for them did not work, resulting in a crash when calling these functions.

This PR fixes https://github.com/Brightify/Cuckoo/issues/533, see that issue for more details.


MatyasKriz:
I'm trying (and failing) to run tests on this branch, do they run fine on your end? I'll have to modify the GH action to run for forked branches as well, so we have CI as the source of truth.

dylanmaryk:
I don't want to spend a huge amount of time trying to fix my local issues compiling the project honestly, so would it be alright with you if I just delete `func withAsyncSendableOptionalClosureAsync` if that works on your end? Would you mind trying deleting that and letting me know if that works?

MatyasKriz:
Oh, it seems like I just forgot to call `make`. You should probably as well, `make` opens the project automatically with everything set up.

It builds and tests correctly. Sorry for the alarm.

In that case I'll merge this and release. 🙂

MatyasKriz:
I have some free time, I'll fix the error and add the missing matchers. 🙂

MatyasKriz:
Merged. 🙂

dylanmaryk:
Thanks, sorry I don't see any changes from you. Do you mean you will fix the error and add the missing matchers later separately?

MatyasKriz:
I don't think I can add commits to your fork, so I've added the commit above your changes (minus the merge commit). 🙂

I've cleaned up the closure matchers file as well, adding missing matcher variations like for optionals and such.

The file could be like 30 lines long if parameter packs worked here. 😕

Either way, it's released with 2.0.15, let me know if it works for you.

dylanmaryk:
Everything does look like it's working on my end, thanks! 🥳

PR #551, opened 2025-04-14
Allow passing an array of values to stub

jdanthinne:
As spreading an array doesn't work in Swift, I've added an extension to be able to pass an array of returned values to a stub.

### DIFF
--- a/Tests/Swift/Stubbing/StubbingTest.swift
+++ b/Tests/Swift/Stubbing/StubbingTest.swift
@@ -15,6 +15,18 @@ final class StubbingTest: XCTestCase {
         XCTAssertEqual(mock.readOnlyProperty, "c")
     }
 
+    func testReturningArray() {
+        let mock = MockTestedClass()
+        stub(mock) { mock in
+            when(mock.readOnlyProperty.get).thenReturn(["a", "b", "c"])
+        }
+
+        XCTAssertEqual(mock.readOnlyProperty, "a")
+        XCTAssertEqual(mock.readOnlyProperty, "b")
+        XCTAssertEqual(mock.readOnlyProperty, "c")
+        XCTAssertEqual(mock.readOnlyProperty, "c")
+    }
+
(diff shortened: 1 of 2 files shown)


Tests/Swift/Stubbing/StubbingTest.swift
```diff
@@ -15,6 +15,18 @@ final class StubbingTest: XCTestCase {
         XCTAssertEqual(mock.readOnlyProperty, "c")
     }
 
+    func testReturningArray() {
+        let mock = MockTestedClass()
+        stub(mock) { mock in
+            when(mock.readOnlyProperty.get).thenReturn(["a", "b", "c"])
```
MatyasKriz, line 21: Won't this be ambiguous when the property is an array?

jdanthinne: Ambiguous for the compiler?
If `readOnlyProperty` is an array, the to generic method will translate to `func thenReturn(_ output: [T], _ outputs: [T]...) -> Self` and `func thenReturn(_ output: [[T]]) -> Self`.

So, if you give it a "simple" array (one dimension), it will known that it is the first method. I you give it an array of arrays (two dimensions), it will know that it is the second one.

Would you prefer to add a name to the argument for the new method, to be clearer for users?

MatyasKriz: Yeah, I'd add something like `inOrder:` to the new method to disambiguate passing array of individually returned values from passing array as a singular one. I can imagine a scenario where some complex tests only tell the developer the classic "The compiler is unable to type-check this expression in reasonable time." because they might not know about the second method which returns the values one by one.

jdanthinne: I've added the `inOrder:` parameter name.

MatyasKriz:
Oh, I forgot to merge this. Merging. 🙂

PR #553, opened 2025-04-15
Fix macOS build

aim2120:
## Description

After Cuckoo recently increased it's range for several dependencies, the package is unable to build for macOS. This is primarily due to the fact that swift-format no longer supports the lowest macOS platform that Cuckoo contains in its Package.swift manifest.

To enable consumers of Cuckoo who are still relying on older macOS versions, in this MR I've introduced a Package.swift with the Swift 5.10 tools version. In this manifest, the range of swift-format and swift-syntax is lowered to be lower than the `600` major version range (in which the macOS platform level `10.15` is still supported).

To confirm this support, I've added a new CI action to test with macOS 14 (which uses Xcode 15).


aim2120:
Hm, Xcode 15 checks passed, but looking at the actual output, the testing was unsuccessful (due to lack of Swift Testing in Xcode 15).

I will push a fix and determine how the CI check was able to pass.

aim2120:
I don't have a macOS 14 machine available to me locally, so it may take some time for me to sort through the macOS 14 CI failures. Will look into it tomorrow, but it's a high priority for me (and others I'm working with) since we really want to update Cuckoo to the latest version.

aim2120:
Testing macOS 14 + Xcode 15 support locally isn't currently possible for me, but I've been running some jobs through CI to try to sus out if the failures observed on this branch may be the result of existing issues in the Cuckoo library.

I was able to verify that building through the Package.swift directly, both on macOS 15 + Xcode 16 and macOS 14 + Xcode 15, the library is currently broken. See this commit and its checks (based off the master branch): https://github.com/aim2120/Cuckoo/commit/47482ceee030f3064fc4f35dbfc3445a8c65ed60

It's obvious that the repo in its current state doesn't support development on macOS 14 + Xcode 14, since it depends upon Swift 6 compiler support (e.g. expecting Swift Testing to be linked automatically). But just to be thorough, this commit (based off the master branch) proves that (again, refer to the checks): https://github.com/aim2120/Cuckoo/commit/c8fe97b2c8b6ee7090849f3c5072ee269455305d

Adding Package.swift build validation to _this_ branch, I can see that there appear to be build errors when building with macOS 14 + Xcode15 (Swift 5 compiler). However, when using macOS 15 + Xcode 16 (Swift 6 compiler), the package builds successfully. See here: https://github.com/aim2120/Cuckoo/commit/56649a492865da617c0fd42b955c3a25d63b9210

Based on these findings, I think it's outside the scope of this MR to ensure support for macOS <15 + Xcode <16. I will be pushing commits to remove the Package.swift for 5.10, remove the backwards compatibility CI checks, and add a Package.swift build check (to ensure the library doesn't get into another state where it's not shippable).

MatyasKriz:
Hey, @aim2120, great PR! Appreciate the help. I'm wondering which change caused this, was is the loosening of dependency requirements in the latest version or something before that?

I've been testing this locally and removing the SwiftFormat dependency seems to fix the problem and since it isn't used in the project yet I'll remove it and add it back when it really is needed, in which case I'll have to raise the minimum required versions.

I think with the SwiftFormat removal it shouldn't be necessary to bump the iOS and macOS versions in this PR, could you please verify and keep only the improvements to the CI process? I'll merge it ASAP.

MatyasKriz:
Oh, my bad, I wanted to rebase the changes above master to test the removal of SwiftFormat, but it created a merge commit instead. Oh well, we'll see what the CI thinks about it.

aim2120:
@MatyasKriz  the issue indeed arose due to the loosening of the version requirements for swift-format, since that libraries does bump the lower bound of supported platforms in later versions.

When running `swift build` on tag `2.0.17`, one gets the following error:
```sh
error: the executable 'CuckooGenerator' requires macos 10.15, but depends on the product 'SwiftFormat' which requires macos 12.0; consider changing the executable 'CuckooGenerator' to require macos 12.0 or later, or the product 'SwiftFormat' to require macos 10.15 or earlier.
```

The incongruity comes from the fact that CI was previously only verified via the Tuist project, which will behave differently from a Package.swift. In fact, on tag `2.0.17`, even if I change the `Generator/Project.swift` `deploymentTargets` to `10.15`, the test target still builds just fine (although, I imagine if I were actually on a 10.15 macOS version, it wouldn't, due to the lack of swift-format support for that version). This appears to be related to how SPM specifically verifies platform version support recursively.

aim2120:
I should note: there is still a gap in CI for Package.swift errors related to _iOS_ version. If you'd like, I could add an additional CI check using `xcodebuild` to verify the Package.swift when built for iOS. If you are happy with just the macOS Package.swift check though, feel free to merge the PR.

PR #565, opened 2025-06-13
Fix typed-throws compilation issue with Swift 6.2

MaelRB:
Newly added typed-throws support doesn't compile with Swift 6.2 for some reason. 
Adding Error protocol requirement on generic ERROR type fixes the issue.

Add a Stub test to assert usage of typed `thenThrow` method.

## Tests
- Xcode 16.4, swift 6.1 ✅
- Xcode 26, swift 6.2 ✅

### DIFF
--- a/Source/Stubbing/Stub.swift
+++ b/Source/Stubbing/Stub.swift
@@ -2,7 +2,7 @@ public protocol Stub {
     var method: String { get }
 }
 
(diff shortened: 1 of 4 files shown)


MatyasKriz:
Thank you for checking that in advance and fixing the issue. 🙂

PR #568, opened 2025-06-23
Update min version for SwiftSyntax to support typed throws

### LINKED ISSUE #567: Cuckoo 2.1.0 build failure (SwiftSyntax)
I've upgraded to 2.1.0 but my project can't build, because `ThrowsClauseSyntax` is not known, as is.
Having checked the SwiftSyntax repo, it seems that it has been introduced in 600.1, so I guess `Package.swift` should be updated to reflect that.

Image

jdanthinne:
Fixes #567

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ Tuist/Dependencies/Cocoapods
 
 Generator/GeneratedMocks.swift
 Generator/Generator.xcodeproj
+.vscode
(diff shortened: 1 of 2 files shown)


MatyasKriz:
Hey, there are some commits part of this PR that shouldn't be. Could you please update the PR to only contain the last commit which updates the version? In that case I can simply merge this, otherwise I have to do it locally and then I'd have to close this PR.

Either way, thanks for the PR! Hopefully it doesn't break things for other folks, though I wonder why CI didn't catch this.

jdanthinne:
Force-pushing broke this PR. Opened this one: https://github.com/Brightify/Cuckoo/pull/569

PR #569, opened 2025-06-30
Update min version for SwiftSyntax to support typed throws

jdanthinne:
Fixes https://github.com/Brightify/Cuckoo/issues/567

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -30,7 +30,7 @@ let package = Package(
         .package(url: "https://github.com/nvzqz/FileKit.git", from: "6.1.0"),
         .package(url: "https://github.com/kylef/Stencil.git", from: "0.15.1"),
         .package(url: "https://github.com/swiftlang/swift-syntax", "509.0.0"..<"603.0.0"),
-        // .package(url: "https://github.com/swiftlang/swift-format", "509.0.0"..<"602.0.0"),
+        // .package(url: "https://github.com/swiftlang/swift-format", "600.1.0"..<"603.0.0"),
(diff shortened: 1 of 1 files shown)


tejassharma96:
if possible, it'd be preferable to wrap the bit that needs this with `#if canImport(SwiftSyntax601)` rather than updating the requirement. not sure if that's possible in this case - if it's something that's completely core to the package then it won't be possible but typed throws sounds like something where we could just `fatalError` in the `#else` case with a message to update the local swift syntax version.

jdanthinne:
I've just looked at the article, but there is no SwiftSyntax601 module in the current version of SwiftSyntax, only 509 and 510. See here: https://github.com/swiftlang/swift-syntax/tree/27db1374d173cb595b52e75a6821bcb6d088873a/Sources

jdanthinne:
Ah, sorry, I just following the link via the PointFree article… but I have to temporarily give up on this as I'm not able to build the project (`make` fails on my machine because it tries to build ruby with openssl1.1, which is now deprecated).

MatyasKriz:
> Ah, sorry, I just following the link via the PointFree article… but I have to temporarily give up on this as I'm not able to build the project (`make` fails on my machine because it tries to build ruby with openssl1.1, which is now deprecated).

That's odd since the version I've defined in Mise is 3.3.8, which is quite recent:
image

MatyasKriz:
I see, looking over the problem I think we'll merge this to make sure compilation errors don't arise and if needed we can expand back to the previous version with the `#if`s implementation.

PR #570, opened 2025-06-30
Update XcodeProj to 9.x

tejassharma96:
Does what the title says 😅 

There weren't any changes to the parts of XcodeProj used in Cuckoo.

If you'd like, I could update this similar to how we have swift-syntax defined (using a version range rather than `from`) but I feel like that's probably overkill for anything except swift-syntax/swift-format

### DIFF
--- a/.package.resolved
+++ b/.package.resolved
@@ -6,8 +6,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/tadija/AEXML.git",
       "state" : {
-        "revision" : "38f7d00b23ecd891e1ee656fa6aeebd6ba04ecc3",
-        "version" : "4.6.1"
+        "revision" : "db806756c989760b35108146381535aec231092b",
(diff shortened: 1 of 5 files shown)


tejassharma96:
hey @MatyasKriz, curious if there's anything blocking this from being released? This is blocking my team's adoption of the latest cuckoo version so I'd love to see it merged. Happy to make it use a range instead of `from` if that would be preferable - whatever it takes to get us up to date on cuckoo!

tejassharma96:
hey @MatyasKriz, apologies for the repeated pings but I'd really love to get my team onto the current cuckoo version and currently am not able to 😓 What would it take to get this merged? happy to make any changes as necessary

tejassharma96:
I don't think that PR is related to this one at all, it should be fine to release just this one. A compatibility fix would certainly be nice but I don't think you'd be causing a regression by releasing this change while figuring out what to do about that issue

Issue #583, opened 2025-11-01
Uncompilable DefaultImpl for generic functions in protocols with associated types


Brennanium:
This is probably a sorta niche use case, but I actually encountered an issue with this in one of my codebases.

A minimal reproduction:
```swift
protocol GenericProtocol {
    associatedtype V
    func genericParameter<T>(value: T)
}
```
currently produces a DefaultImpl class that doesn't compile:
```swift
...
class DefaultImplCaller: GenericProtocol, @unchecked Sendable {
    private let reference: Any
    init<_CUCKOO$$GENERIC: GenericProtocol>(from defaultImpl: UnsafeMutablePointer<_CUCKOO$$GENERIC>, keeping reference: @escaping @autoclosure () -> Any?) where _CUCKOO$$GENERIC.V == V {
        self.reference = reference
    
        _storage$1$genericParameter = defaultImpl.pointee.genericParameter // error: Generic parameter 'T' could not be inferred
    }

    private let _storage$1$genericParameter: (T) -> Void // error: Cannot find type 'T' in scope
    func genericParameter<T> (value p0: T) {
        return _storage$1$genericParameter(p0)
    }
}
...
```

I had to fight with the compiler a bit, but I managed to figure out a fix.  I'll be opening up a PR soon, but I just wanted to open up an issue to link to first.

PR #584, opened 2025-11-01
Fix protocol DefaultImpl with associated types + generic functions

Brennanium:
Fix for issue #583 by both constraining and opening the existential type and using stored `reference` instead of storage for generic functions

Making `reference` a non-optional constrained existential instead of `Any?` type seems to work for all of the test cases I could come up with, and keeping the `pointee` in the autoclosure as the `reference` in the `enableDefaultImplementation<>(mutating:)` init could possibly cause a retain cycle or something?  But since we're already capturing the UnsafeMutablePointer for getters and setters anyways, it seemed like an ok thing try.

### DIFF
--- a/Generator/Sources/Internal/Templates/MockTemplate.swift
+++ b/Generator/Sources/Internal/Templates/MockTemplate.swift
@@ -16,7 +16,8 @@ extension Templates {
 extension {{ container.parentFullyQualifiedName }} {
 {% endif %}
 {% if container.hasPrimaryAssociatedTypes %}
-@available(iOS 16.0.0, macOS 13.0.0, watchOS 9.0, tvOS 16, *) // runtime support for constrained protocols with primary associated types
+// runtime support for constrained protocols with primary associated types
+@available(iOS 16, macOS 13, watchOS 9, tvOS 16, *)
 {% endif %}
 {{ container.accessibility|withSpace }}class {{ container.mockName }}{{ container.genericParameters }}:{% if container.isNSObjectProtocol %} NSObject,{% endif %} {{ container.name }}{% if container.isImplementation %}{{ container.genericArguments }}{% endif %},{% if container.isImplementation %} Cuckoo.ClassMock{% else %} Cuckoo.ProtocolMock{% endif %}, @unchecked Sendable {
     {% if container.isGeneric and not container.isImplementation and not container.hasOnlyPrimaryAssociatedTypes %}
--- a/Generator/Sources/Internal/Tokens/Capabilities/HasGenerics.swift
+++ b/Generator/Sources/Internal/Tokens/Capabilities/HasGenerics.swift
@@ -31,15 +31,17 @@ extension HasGenerics {
     }
 
     func genericsSerialize() -> GeneratorContext {
-        var genericProtocolIdentity: String?
-        var genericPrimaryAssociatedTypeArguments: String?
-        
-        if let protocolDeclaration = asProtocol {
-            genericProtocolIdentity = genericParameters.map { "\(Templates.staticGenericParameter).\($0.name) == \($0.name)" }.joined(separator: ", ")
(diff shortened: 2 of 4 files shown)


Brennanium:
Given this example protocol:
```swift
protocol GenericProtocol<C> {
    associatedtype C
    associatedtype V
    func genericParameter<T>(value: T)
    func genericParameterWithAssociatedType<T>(value: T, theC: C, theV: V) -> V
}
```
Here's a diff of how it would generate before (with compile errors) vs how it generates now:
```diff
...
class DefaultImplCaller: GenericProtocol, @unchecked Sendable {
-    private let reference: Any
+    private let reference: () -> any GenericProtocol<C>

-    init<_CUCKOO$$GENERIC: GenericProtocol>(from defaultImpl: UnsafeMutablePointer<_CUCKOO$$GENERIC>, keeping reference: @escaping @autoclosure () -> Any?) where _CUCKOO$$GENERIC.C == C, _CUCKOO$$GENERIC.V == V {
+    init<_CUCKOO$$GENERIC: GenericProtocol>(from defaultImpl: UnsafeMutablePointer<_CUCKOO$$GENERIC>, keeping reference: @escaping @autoclosure () -> _CUCKOO$$GENERIC) where _CUCKOO$$GENERIC.C == C, _CUCKOO$$GENERIC.V == V {
        self.reference = reference
    
-        _storage$1$genericParameter = defaultImpl.pointee.genericParameter // error: Generic parameter 'T' could not be inferred
-        _storage$2$genericParameterWithAssociatedType = defaultImpl.pointee.genericParameterWithAssociatedType // error: Generic parameter 'T' could not be inferred
    }

-    private let _storage$1$genericParameter: (T) -> Void // error: Cannot find type 'T' in scope
    func genericParameter<T> (value p0: T) {
-        return _storage$1$genericParameter(p0)
+.       return reference().genericParameter(p0) 
    }
-    private let _storage$2$genericParameterWithAssociatedType: (T, C, V) -> V // error: Cannot find type 'T' in scope
    func genericParameterWithAssociatedType <T> (value p0: T, theC p1: C, theV p2: V) -> V {
+        func openExistential<_CUCKOO$$GENERIC: GenericProtocol<C>>(_ opened: _CUCKOO$$GENERIC) -> V {
+            return opened.genericAndAassociatedTypeParameters(value: p0, theC: p1, theV: p2 as! _CUCKOO$$GENERIC.V) as! V
+        }
+        return openExistential(reference())
-        return _storage$2$genericParameterWithAssociatedType(p0)
    }
}
...
```

Generator/Sources/Internal/Templates/MockTemplate.swift
```diff
@@ -15,11 +15,16 @@ extension Templates {
 {% if container.hasParent %}
 extension {{ container.parentFullyQualifiedName }} {
 {% endif %}
+{% if container.hasPrimaryAssociatedTypes %}
+@available(iOS 16.0.0, macOS 13.0.0, watchOS 9.0, tvOS 16, *) // runtime support for constrained protocols with primary associated types
```
MatyasKriz:
```suggestion
// runtime support for constrained protocols with primary associated types
@available(iOS 16, macOS 13, watchOS 9, tvOS 16, *)
```

Brennanium:
@MatyasKriz Glad to help! 😊 I've addressed all your nitpick suggestions in a follow-up commit; let me know if you encounter any other issues when you manage to run it locally.

Brennanium:
@via-guy I looked into what it would take to get mocks generated for protocols with same-type requirements working, and unfortunately it appears to mostly be unrelated to the changes I'm making here in this PR.  Tackling the problem would likely involve beefing up the `Token.mergingInheritance(with:)` logic which currently traverses all the inherited types only to collect the member tokens.   You'd need to add some new complex behavior merging and diffing all of the inherited associated types + primary associated types and checking them against all the same-type requirements.  It's probably doable, but I'd say out of the scope of this specific issue.

PR #585, opened 2025-11-07
fix(podspec): remove sub.source_files = []

veraposeidon:
remove sub.source_files = [] which caused OCMock subspec to fetch no source files


MatyasKriz:
Ah, thanks for the catch!

PR #588, opened 2025-11-25
Expose CuckooGenerator as executable product

nmcc24:
## Summary
Exposes `CuckooGenerator` as a public executable product so downstream consumers can build
custom build plugins on top of it. In order to do so, `XcodeProj` must be bumped to 9.13.0+. 

## Changes

| File | Change |
|---|---|
| `Package.swift` | Adds `.executable(name: "CuckooGenerator", targets: ["CuckooGenerator"])` product; bumps `XcodeProj` `9.9.0` → `9.14.0`; raises `.macOS` platform floor `10.15` → `11.0` |
| `Package.resolved` | Re-resolved pin for `XcodeProj` → `9.14.0` (`eb001108...`), `originHash` regenerated |
| `Generator/Sources/Internal/Helpers/Async+convenience.swift` | `#available(macOS 14.0, *)` → `#available(macOS 14.0, iOS 17.0, *)` — adds the `iOS` branch needed now that `CuckooGenerator` can be built for iOS as a side effect of the executable product |

## Verification
- `swift package resolve` — resolves cleanly against the new `XcodeProj` version.
- `swift build` — full build succeeds (`CuckooGenerator` and `CuckooGenerator-tool` link successfully).

## Things to call out to reviewers
- **Breaking for consumers still on macOS 10.15**: the platform bump to macOS 11.0 is required
  by `XcodeProj` 9.14.0 itself (its manifest declares 11.0 as its floor), not a choice made here.
  Worth a one-line mention in the PR body since it's a compatibility change, not just a version bump.
- **iOS compilation of `CuckooGenerator`**: per the earlier PR discussion, exposing this as an
  executable product means it now also compiles for iOS as a side effect of the platform list,
  even though it's really meant for host-side (macOS) tooling use. Nothing in `Package.swift`
  stops an app target from linking it. Confirm the README documents that this is intended for
  build-tool/plugin use only, not for shipping inside an app target.
- No behavior changes to existing `Cuckoo`/mocking runtime code — this is purely packaging +
  dependency surface.

### DIFF
--- a/Generator/Sources/Internal/Helpers/Async+convenience.swift
+++ b/Generator/Sources/Internal/Helpers/Async+convenience.swift
@@ -47,7 +47,7 @@ extension Sequence where Element: Sendable {
         // A task group automatically waits for all of its
         // sub-tasks to complete, while also performing those
         // tasks in parallel:
-        if #available(macOS 14.0, *) {
+        if #available(macOS 14.0, iOS 17.0, *) {
             await withDiscardingTaskGroup { group in
                 for element in self {
                     group.addTask {
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "56435a5ee34a0880cd334d8501491902482701af75e23031558e8d751a0ff571",
+  "originHash" : "d74f646b284cb356ac99bcf2d5948e8a933f89b92eccac29090027d4bdc94411",
   "pins" : [
     {
       "identity" : "aexml",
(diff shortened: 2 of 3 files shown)


nmcc24:
Hi @MatyasKriz 

Thanks for your response and sharing your views. 

The compilation for iOS is simply a side effect of running the build tool plugging when targeting that platform. 

As far as I have observed, there is no way of restricting a package (and its products) to a particular platform. Even if one does not mentioned them in the Package.swift file, it will end up defaulting to a minimum version. 

It is true that by opening up the executable as public product would allow that use case to occur. However, it seems to me that it wouldn't ever work on a platform other than MacOS/Linux. By complementing this with a potential mention on the README file, I believe we would be providing any consumer with enough information - I appreciate however this is very subjective :)

In short, unless we consider modifying the plugin you provide (which I don't think we should consider as it would introduce unnecessary and niche complexity to this repo), I can't see any other way that allows us to enable the composition I'm trying to achieve.

nmcc24:
@MatyasKriz this is now finally ready to be reviewed as XcodeProj 9.13.0+ now has all necessary changes required to make this happen.

MatyasKriz (approved):
Looks good, merging. We'll revisit this if users find themselves confused when adding the library.

Issue #555, opened 2025-04-29
Multiple modules in Cuckoo toml file


adelamarre-pelm:
In the toml file, I noticed there is the ability to have multiple modules in the toml file.  Is this feature working? 

In my toml file I've added multiple modules, specifying two different output file names.  However when build my test target, only one generated mocks file is created in Derived data, in addition the name of the generated mocks file doesn't change, it remains GeneratedMocks.swift.

MatyasKriz:
Hi, yes, the feature works, but unfortunately the SPM build plugin was quite limited in its functionality, so I admit it's not completely up to snuff.

If you can, run the cuckoonator from a build script without sandboxing, that way there's no limitations and the outputs are respected.

I have a lot on my plate at the moment, but if I have some time I'll see if this process can be improved.

PR #595, opened 2026-03-11
Add CuckooPluginModular for per-module mock generation

SylvainRX:
### Summary
Adds a new `CuckooPluginModular` build tool plugin that generates separate mock files per module dependency, improving modularity for multi-target Swift Packages.

Related issue: https://github.com/Brightify/Cuckoo/issues/555

### Motivation
The existing `CuckooPluginSingleFile` generates all mocks into a single `GeneratedMocks.swift` file in derived data. In a Swift Package with multiple targets, each test target compiles independently and may not have visibility into types from unrelated modules, making a single shared mock file problematic.

#### New Plugin: CuckooPluginModular

- Located in `Generator/Plugin/Modular/`
- Inspects the test target's dependencies (excluding Cuckoo) and runs the generator once per dependency module
- Each module gets its own build command with `CUCKOO_MODULE_NAME` environment variable
- Generates one mock file per module: `GeneratedMocks_<ModuleName>.swift`
- Also emits a build command keyed by the test target's own name, aggregating all dependency sources — this allows `Cuckoofile.toml` to have a `[modules.<TestTargetName>]` entry to control which files are mocked and which imports are added
- A `Cuckoofile.toml` entry is required for each module; modules without a matching entry produce an empty file

### Usage

**Package.swift:**
```swift
.testTarget(
    name: "TargetATests",
    dependencies: ["TargetA", "Cuckoo"],
    plugins: [
        .plugin(name: "CuckooPluginModular", package: "Cuckoo"),
    ]
),
.testTarget(
    name: "AggregationTargetTests",
    dependencies: ["AggregationTarget", "Cuckoo"],
    plugins: [
        .plugin(name: "CuckooPluginModular", package: "Cuckoo"),
    ]
)
```

**Cuckoofile.toml:**
```toml
# TargetA mocks
[modules.TargetATests]
imports = ["Foundation"]
testableImports = ["TargetA"]
sources = [
    "Sources/TargetA/InternalProtocolA.swift"
]

# AggregationTarget mocks - multiple imports
[modules.AggregationTargetTests]
imports = ["Foundation", "TargetA", "TargetB"]
sources = [
    "Sources/TargetA/ProtocolA.swift",
    "Sources/TargetB/ProtocolB.swift",
]
```

### Breaking Changes
None. All changes are backward compatible with existing plugins and configurations.

### DIFF
--- a/Generator/Sources/CLI/GenerateCommand.swift
+++ b/Generator/Sources/CLI/GenerateCommand.swift
@@ -56,11 +56,9 @@ struct GenerateCommand: AsyncParsableCommand {
         )
 
         if modules.isEmpty {
-            let compoundModuleName = ProcessInfo.processInfo.environment["CUCKOO_COMPOUND_MODULE_NAME"]
             let requestedModuleName = ProcessInfo.processInfo.environment["CUCKOO_MODULE_NAME"]
-            let effectiveName = compoundModuleName ?? requestedModuleName
-            if let effectiveName {
-                log(.info, message: "No module named '\(effectiveName)' found in Cuckoofile, skipping generation.")
+            if let requestedModuleName {
+                log(.info, message: "No module named '\(requestedModuleName)' found in Cuckoofile, skipping generation.")
             }
             if let outputPath = overriddenOutput {
                 let path = Path(outputPath, expandingTilde: true)
@@ -124,7 +122,6 @@ struct GenerateCommand: AsyncParsableCommand {
 
     func modules(configurationPath: Path, contents: String) throws -> [Module] {
         let requestedModuleName = ProcessInfo.processInfo.environment["CUCKOO_MODULE_NAME"]
-        let compoundModuleName = ProcessInfo.processInfo.environment["CUCKOO_COMPOUND_MODULE_NAME"]
 
         var errorMessages: [String] = []
         var globalOutput: String? = overriddenOutput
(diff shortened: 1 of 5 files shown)


Generator/Sources/CLI/GenerateCommand.swift
```diff
@@ -50,11 +50,49 @@ struct GenerateCommand: AsyncParsableCommand {
...
+                // No compound override – fall back to the plain module name.
+                modules = allModules.filter { $0.name == requestedModuleName }
+            } else {
+                modules = []
+            }
+        } else if let requestedModuleName {
+            modules = allModules.filter { $0.name == requestedModuleName }
+        } else {
+            modules = allModules
+        }
```
MatyasKriz: Since the `allModules` aren't used (and I can't think of why they would be), I'd like this filtering logic moved there instead, keeping this high-level implementation easy to read.

Package.swift
```diff
@@ -19,6 +19,10 @@ let package = Package(
             name: "CuckooPluginSingleFile",
             targets: ["CuckooPluginSingleFile"]
         ),
+        .plugin(
+            name: "CuckooPluginPerModule",
```
MatyasKriz: I'd use "CuckooPluginModular" instead, does that make sense for this implementation?

Generator/Sources/CLI/GenerateCommand.swift
MatyasKriz, line 132: I'm not sure comparing potentially large strings on every generation run is the way to go here. I'd rather cache the mocked file names and their latest modification date which is then compared to the generated file modification date. I'm not asking you to implement that, buut I'd rather leave the logic as it is until that is implemented.

MatyasKriz (commented):
Looks great! I appreciate you taking the time to improve the project like this. Especially the automatic source detection might be nice for small projects.

One more thing, though – please mention this new modular build tool in README, so that users don't miss it. 🙂

SylvainRX:
@MatyasKriz Thank you for the review and feedbacks! I'll address them as soon as I have some time.

Package.swift
SylvainRX: It does make sense, done ✅

Generator/Sources/CLI/GenerateCommand.swift
SylvainRX: I ended up getting rid of the concept of compound modules altogether, I don't think it was bringing that much value compared to what I have now. I updated the PR description accordingly

Generator/Sources/CLI/GenerateCommand.swift
SylvainRX: Right, I reverted this part above

SylvainRX (commented):
Hello @MatyasKriz ! 
I pushed changes addressing your comments, and updated the readme as well. 
Please let me know if there's thing I can keep improving on.
Thanks!

MatyasKriz (approved):
Sorry for the delay, looks good! The documentation is a nice touch as well. 👏

SylvainRX:
Thank you @MatyasKriz !

PR #602, opened 2026-08-08
Fix generator version mismatch

### LINKED ISSUE #601: Cuckoo Generator version does not match Swift Package version.
The cuckoo generator version is stuck at 2.0.10 while the Swift Package version is 2.4
https://github.com/Brightify/Cuckoo/blob/master/Generator/Sources/CLI/Version.swift

ZhangStudyLife:
## Summary

- update the Cuckoo generator version from `2.0.10` to `2.4.0`
- add a regression test that keeps the generator version aligned with the root `version` file

## Testing

- verified the generator version matches the root `version` file
- `git diff --check`
- macOS Swift/Xcode tests were not run locally because this environment is Windows

Closes #601

### DIFF
--- a/Generator/Sources/CLI/Version.swift
+++ b/Generator/Sources/CLI/Version.swift
@@ -1,1 +1,1 @@
-let version = "2.0.10"
+let version = "2.4.0"
--- a/Generator/Tests/VersionTest.swift
+++ b/Generator/Tests/VersionTest.swift
@@ -1,0 +1,17 @@
+import Foundation
+import XCTest
+
+final class VersionTest: XCTestCase {
+    func testGeneratorVersionMatchesLibraryVersion() throws {
+        let repositoryURL = URL(fileURLWithPath: #filePath)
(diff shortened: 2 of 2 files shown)


ZhangStudyLife:
Update after commit `d723d71`: the latest [Run tests workflow](https://github.com/Brightify/Cuckoo/actions/runs/31819242861) is still `action_required` and contains zero jobs, so no test has failed yet—the workflow has not been approved to start.

The PR now explicitly runs the existing `Generator` scheme, whose test action includes `CuckoonatorTests` and the new version regression test. The workflow YAML parses successfully. I cannot execute the Xcode test locally on this Windows/WSL machine because Swift/Xcode is not installed.

Could a maintainer please approve the workflow and review the change when convenient? Thanks!

MatyasKriz:
Thank you for your contribution, I ran the workflows shortly after your ping, but forgot to check and it seems like it broke generator tests. Take a look.

